PR #23, opened 2018-11-22
precompile typescript and run javascript in production

wrboyce:
* core: Precompile typescript and run javascript in production
* core: Update npm dependencies
* core: Run unit tests when building docker image

Connects-to: #19
Change-type: minor

---

##### Contributor checklist

- [X] Tests are included
- [X] Affects the development, build or deployment processes of the component

##### Reviewer Guidelines
---


src/utils/index.ts
```diff
@@ -15,21 +15,21 @@
...
-} from 'pinejs-client-request';
-export { PinejsClientCoreFactory } from 'pinejs-client-request';
-import * as pkg from 'pjson';
 import * as winston from 'winston';
 
-export type AnyObject = PinejsClientCoreFactory.AnyObject;
+import { PinejsClientRequest } from 'pinejs-client-request';
+export { PinejsClientCoreFactory } from 'pinejs-client-request';
+
+import { version as VERSION } from '../../package.json';
```
Page-: This can be
```suggestion
export { version as VERSION } from '../../package.json';

File: src/utils/index.ts [hunk 2]
```diff
@@ -15,21 +15,21 @@
...
+export { VERSION };
+
+import * as errors from './errors';
+export { errors };
```
Page-: I think it'd make sense for other modules to import `utils/error` instead, I don't see a reason to alias internal modules (the main one I'm aware of is for the case of exposing a single entrypoint for a node module, but that doesn't apply here)

Dockerfile
```diff
@@ -25,7 +25,16 @@ RUN echo "AUTOSTART=none" > /etc/default/openvpn \
...
 COPY . /usr/src/app
-RUN npm run check
 
 COPY config/services /etc/systemd/system
-RUN systemctl enable open-balena-vpn.service open-balena-connect-proxy.service
+RUN systemctl enable open-balena-vpn-api.service open-balena-connect-proxy.service
+
+# build test image
+FROM base as test
+RUN npm install && npm cache clean --force 2>/dev/null
```
Page-: This should be
```suggestion
RUN npm ci && npm cache clean --force 2>/dev/null

File: Dockerfile [hunk 4]
```diff
@@ -25,7 +25,16 @@ RUN echo "AUTOSTART=none" > /etc/default/openvpn \
 COPY package.json package-lock.json /usr/src/app/
 RUN npm install --unsafe-perm --production && npm cache clean --force 2>/dev/null
```
Page-: This should be:
```suggestion
RUN npm ci --unsafe-perm --production && npm cache clean --force 2>/dev/null

File: Dockerfile (hunk 3 above)
Comment (wrboyce) [reply Dockerfile]: Why does it need to be a clean install? It is just adding the development packages so the unit tests can be run.

File: Dockerfile (hunk 4 above)
Comment (wrboyce) [reply Dockerfile]: Is the install not already clean by virtue of being in a docker image? We've only added the `package{,-lock}.json` files by this point.

File: src/vpn-api/worker.ts [hunk 5]
```diff
@@ -22,9 +22,11 @@ import * as forever from 'forever-monitor';
 import * as morgan from 'morgan';
 import * as net from 'net';
 
+import { logger } from '../utils';
+import * as errors from '../utils/errors';
```
Page-: If you change this (and similar places) back to
```ts
import { captureException, Raven } from './utils/errors';
```
then it'll make for a much smaller diff

src/vpn-api/utils/index.ts
```diff
@@ -15,9 +15,8 @@
...
-
-export const request = rp.defaults({
-	resolveWithFullResponse: true,
-	simple: false,
-});
+import * as clients from './clients';
+export { clients };
+export { Netmask } from './netmask';
+export { getPostWorker, request } from './request';
+export { service, ServiceInstance } from './service';
```
Page-: Are all of these exports still used?

test/connect-proxy/device.ts
```diff
@@ -26,7 +26,9 @@ const { expect } = chai;
 nock.disableNetConnect();
 
 const BALENA_API_HOST = process.env.BALENA_API_HOST!;
-const VPN_SERVICE_API_KEY = Buffer.from(process.env.VPN_SERVICE_API_KEY!);
+const VPN_SERVICE_API_KEY = Buffer.from(
+	process.env.VPN_SERVICE_API_KEY || 'test_vpn_string',
```
dfunckt: I think it's better to fail if the env var is not set than falling back to an arbitrary default that only really makes sense in tests. The way it is now, it may mask misconfiguration.

bin/entry.sh
```diff
@@ -0,0 +1,27 @@
...
+set -eu
+
+WORK_DIR="$(readlink -e "$(dirname "$0")"/..)"
+PATH="${WORK_DIR}/node_modules/.bin:${PATH}"
+cd "${WORK_DIR}"
+
+test -f "${WORK_DIR}/config/env" && source "${WORK_DIR}/config/env"
+
+service="$(basename "$0")"
+service="${service/balena-/}"
```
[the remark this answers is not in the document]
wrboyce: I was just following the pattern I use in the proxy.

test/connect-proxy/device.ts
wrboyce: I'm not sure I follow, this is only for tests which will be run under different config environments anyway. Adding the default here was to reduce the required `ENV` lines in the `Dockerfile` when running the tests when building (which itself is a workaround so the CI runs the tests).

dfunckt: Oh, yeah, I didn't realise that file was a test file :)

PR #29, opened 2019-03-04
Maintenance Fixes/Updates

wrboyce:
Change-type: minor

---
##### Contributor checklist

- [ ] Tests are included
- [X] Affects the development, build or deployment processes of the component

##### Reviewer Guidelines
---

### DIFF
--- a/bin/entry.sh
+++ b/bin/entry.sh
@@ -5,16 +5,19 @@ WORK_DIR="$(readlink -e "$(dirname "$0")"/..)"
 PATH="${WORK_DIR}/node_modules/.bin:${PATH}"
 cd "${WORK_DIR}"
 
+# shellcheck disable=1090
 test -f "${WORK_DIR}/config/env" && source "${WORK_DIR}/config/env"
 
 service="$(basename "$0")"
 service="${service/balena-/}"
 command="$(command -v node)"
+args=("-r" "source-map-support/register")
 entrypoint="build/src/${service}/app.js"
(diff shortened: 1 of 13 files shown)


bin/entry.sh
```diff
@@ -5,16 +5,19 @@ WORK_DIR="$(readlink -e "$(dirname "$0")"/..)"
 PATH="${WORK_DIR}/node_modules/.bin:${PATH}"
 cd "${WORK_DIR}"
 
+# shellcheck disable=1090
 test -f "${WORK_DIR}/config/env" && source "${WORK_DIR}/config/env"
 
 service="$(basename "$0")"
 service="${service/balena-/}"
 command="$(command -v node)"
+args=("-r" "source-map-support/register")
```
Page-: Is this for nicer sentry errors?  It might be better in that case to upload the source maps to sentry and let it handle the mapping rather than paying the cost in production

wrboyce: I was actually thinking of nicer local errors, and hadn't even considered Sentry. I can certainly see the argument for "run typescript locally, provide Sentry with source maps" though -- I'll look into it!

wrboyce: While getting the source maps into Sentry is definitely the best way forward, it seems to be no small undertaking so I'm going to merge as is for now until we can get Concourse handling Sentry releases properly.

Page-: Ok, I'm just wondering if this undoes the benefits from not transpiling typescript

PR #32, opened 2019-03-08
vpn: collect per-device bandwidth stats & publish for prometheus

wrboyce:
Connects-to: #31
Change-type: minor
Depends-on: balena-io/resin-containers#589

---

##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---

### DIFF
--- a/config/confd/conf.d/haproxy.toml
+++ b/config/confd/conf.d/haproxy.toml
@@ -1,4 +1,7 @@
 [template]
 src = "haproxy.cfg.tmpl"
 dest = "/etc/haproxy/haproxy.cfg"
-keys = [ "/resin/vpn/" ]
+keys = [
+    "/balena/monitor/secret-token",
+    "/resin/vpn/"
+]
--- a/config/confd/templates/haproxy.cfg.tmpl
+++ b/config/confd/templates/haproxy.cfg.tmpl
@@ -17,7 +17,6 @@ global
 	stats socket /var/run/haproxy.sock mode 600 level admin
 
 defaults
-	mode tcp
 	maxconn {{$maxconn}}
 	log global
 	option dontlognull
@@ -29,21 +28,38 @@ defaults
 	timeout client {{$timeout}}s
 	timeout server {{$timeout}}s
 
+{{ if ne (getv "/balena/monitor/secret-token" "") "" -}}
+userlist metrics
+	user monitor insecure-password "{{ getv "/balena/monitor/secret-token" }}"
+{{- end }}
+
 frontend health-80
 	mode http
 	option httplog
 	bind ipv4@:80
 	bind ipv4@:81 accept-proxy
-	acl vpn_dead nbsrv(vpn-cluster) lt 1
+	acl vpn_dead nbsrv(vpn-workers) lt 1
 	monitor-uri /ping
 	monitor fail if vpn_dead
+{{- if ne (getv "/balena/monitor/secret-token" "") "" }}
+	acl is-metrics path /metrics
+	acl metrics-auth http_auth(metrics)
+	http-request deny if is-metrics !metrics-auth
+	use_backend vpn-master if is-metrics metrics-auth
+
+backend vpn-master
+	mode http
+	server vpn0 127.0.0.1:8080 check
+{{- end }}
 
 frontend tcp-{{$vpnPort}}
+	mode tcp
 	bind ipv4@:{{$vpnPort}} {{$bindOpt}}
 	# Routing <client-ip:port>@<frontend> to <backend>/<server>:<port> [Conns Queues Times]
 	log-format "Routing %ci:%cp@%ft to %b/%s:%bp [C:%bc/%sc Q:%bq/%sq T:%Tw/%Tc]"
-	default_backend vpn-cluster
+	default_backend vpn-workers
 
-backend vpn-cluster
+backend vpn-workers
+	mode tcp
 	balance leastconn
 	server-template vpn 1-{{$servers}} 127.0.0.1:10000 check disabled
--- a/src/vpn-api/app.ts
+++ b/src/vpn-api/app.ts
@@ -16,8 +16,12 @@
 */
 
 import * as cluster from 'cluster';
+import * as express from 'express';
 import * as _ from 'lodash';
+import * as morgan from 'morgan';
 import * as os from 'os';
+
+import { metrics } from '@balena/node-metrics-gatherer';
 
 import { logger, VERSION } from '../utils';
 
@@ -40,29 +44,34 @@ if (cluster.isMaster) {
 	logger.info(
 		`open-balena-vpn@${VERSION} master process started with pid ${process.pid}`,
 	);
-	if (VPN_INSTANCE_COUNT > 1) {
-		logger.info(`spawning ${VPN_INSTANCE_COUNT} workers`);
-		_.times(VPN_INSTANCE_COUNT, i => {
-			const instanceId = i + 1;
-			const restartWorker = (code?: number, signal?: string) => {
-				if (signal != null) {
-					logger.error(
-						`open-balena-vpn worker-${instanceId} killed with signal ${signal}`,
-					);
-				}
-				if (code != null) {
-					logger.error(
-						`open-balena-vpn worker-${instanceId} exited with code ${code}`,
-					);
-				}
-				cluster.fork({ VPN_INSTANCE_ID: instanceId }).on('exit', restartWorker);
-			};
-			restartWorker();
-		});
-	}
+	logger.info(`spawning ${VPN_INSTANCE_COUNT} workers`);
+	_.times(VPN_INSTANCE_COUNT, i => {
+		const instanceId = i + 1;
+		const restartWorker = (code?: number, signal?: string) => {
+			if (signal != null) {
+				logger.error(
+					`open-balena-vpn worker-${instanceId} killed with signal ${signal}`,
+				);
+			}
+			if (code != null) {
+				logger.error(
+					`open-balena-vpn worker-${instanceId} exited with code ${code}`,
+				);
+			}
+			cluster.fork({ VPN_INSTANCE_ID: instanceId }).on('exit', restartWorker);
+		};
+		restartWorker();
+	});
+
+	const app = express();
+	app.disable('x-powered-by');
+	app.get('/ping', (_req, res) => res.send('OK'));
+	app.use(morgan('combined'));
+	app.get('/metrics', metrics.aggregateRequestHandler());
+	app.listen(8080);
 }
 
-if (cluster.isWorker || VPN_INSTANCE_COUNT === 1) {
-	const instanceId = parseInt(process.env.VPN_INSTANCE_ID || '1', 10);
+if (cluster.isWorker) {
+	const instanceId = parseInt(process.env.VPN_INSTANCE_ID!, 10);
 	service.wrap(() => worker(instanceId));
 }
--- a/src/vpn-api/worker.ts
+++ b/src/vpn-api/worker.ts
@@ -19,8 +19,12 @@ import * as Promise from 'bluebird';
 import * as compression from 'compression';
 import * as express from 'express';
 import * as forever from 'forever-monitor';
+import * as _ from 'lodash';
 import * as morgan from 'morgan';
 import * as net from 'net';
+import VpnManager = require('telnet-openvpn');
+
+import { metrics } from '@balena/node-metrics-gatherer';
 
 import { captureException, logger } from '../utils';
 import { Raven } from '../utils/errors';
@@ -28,7 +32,7 @@ import { Raven } from '../utils/errors';
 import apiFactory from './api';
 import { Netmask } from './utils';
 
-interface AsyncApplication extends express.Application {
+interface AsyncApplication extends express.Express {
 	listenAsync(port: number): Promise<ReturnType<express.Application['listen']>>;
 }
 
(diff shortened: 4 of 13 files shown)


src/vpn-api/worker.ts
```diff
@@ -143,13 +162,77 @@ const worker = (instanceId: number) => {
...
+					if (clients[clientId] == null) {
+						logger.error(`Unknown CID(${clientId}) from OpenVPN!`);
+						return;
+					}
+
+					const { uuid } = clients[clientId];
+					const ts = new Date().getTime();
+					const rxBytes = parseInt(data.bytecount[1], 10);
+					const txBytes = parseInt(data.bytecount[2], 10);
+					if (process != null && process.send != null) {
```
Page-: Can `process` be null if we're running in nodejs?  What about `process.send`, when would that be null?

src/vpn-api/worker.ts
```diff
@@ -143,13 +162,77 @@ const worker = (instanceId: number) => {
...
+						};
+						vpn.on('log', idMapper);
+					}
+				});
+				// process bytecount events to track realtime data usage
+				vpn.on('data', data => {
+					if (data.bytecount == null) {
+						return;
+					}
+					if (!data.bytecount[0].includes('CLI:')) {
```
Page-: Is the `CLI:` guaranteed to be at the start?  I'm wondering because you split on `:` and take the second element below so I'm expecting it to be to extract this.  If so it might be better to do the split and do an `===` on the first element

src/vpn-api/worker.ts
```diff
@@ -143,13 +162,77 @@ const worker = (instanceId: number) => {
...
+		.tap(() => {
+			return vpn.connect({ port: mgtPort, shellPrompt: '' }).then(() => {
+				vpn.connection.shellPrompt = '';
+				// monitor new client connections and map cid to uuid
+				vpn.on('log', data => {
+					if (data.includes('CLIENT:ESTABLISHED,')) {
+						const clientId = data.split(',')[1].trim();
+						const idMapper = (logData: string) => {
+							if (logData.includes('CLIENT:ENV,common_name=')) {
+								const now = new Date().getTime();
```
Page-:
```suggestion
								const now = Date.now();
```

src/vpn-api/app.ts
```diff
@@ -35,34 +39,121 @@ import worker from './worker';
...
+	app.disable('x-powered-by');
+	app.get('/ping', (_req, res) => res.send('OK'));
+	app.use(morgan('combined'));
+	app.get('/metrics', (_req, res) => {
+		// res.writeHead(200, { 'Content-Type': 'text/plain' });
+		_.map(
+			_.filter(
+				sessions,
+				session =>
+					new Date().getTime() - session.ts <
```
Page-:
```suggestion
					Date.now() - session.ts <
```

src/vpn-api/app.ts
```diff
@@ -35,34 +39,121 @@ import worker from './worker';
...
+					VPN_BYTECOUNT_INTERVAL * 1000 * 1.5,
```
Page-: What is this magic?  Also it'd be nice to calculate it just the once and assign to a named var

src/vpn-api/app.ts
Page-: Can you grab the time just once rather than each loop (it's a surprisingly expensive operation).  You can also do it as:
```ts
const x = Date.now() - VPN_BYTECOUNT_INTERVAL * 1000 * 1.5
_.filter(sessions, session => session.ts > x)

File: src/vpn-api/app.ts [hunk 6]
```diff
@@ -35,34 +39,121 @@ import worker from './worker';
...
+			labelNames: [],
+		},
+	);
+
+	const app = express();
+	app.disable('x-powered-by');
+	app.get('/ping', (_req, res) => res.send('OK'));
+	app.use(morgan('combined'));
+	app.get('/metrics', (_req, res) => {
+		// res.writeHead(200, { 'Content-Type': 'text/plain' });
```
Page-: debug?

src/vpn-api/app.ts
```diff
@@ -35,34 +39,121 @@ import worker from './worker';
...
+				500 * 1024 * 1024,
+				1024 * 1024 * 1024,
+				10 * 1024 * 1024 * 1024,
+				50 * 1024 * 1024 * 1024,
+				100 * 1024 * 1024 * 1024,
+				150 * 1024 * 1024 * 1024,
+				200 * 1024 * 1024 * 1024,
+				250 * 1024 * 1024 * 1024,
+				500 * 1024 * 1024 * 1024,
+				1024 * 1024 * 1024 * 1024,
```
Page-: Could you use vars for this to make it more readable, ie
```ts
500 * GB
1 * TB,
```

src/vpn-api/worker.ts
wrboyce: Yeah I was pretty surprised at this one myself, but typescript complained that `process.send` might be null… The check on `process` is probably extraneous though.

Page-: It makes sense, it can be undefined if you're the master or a child without an IPC channel: https://nodejs.org/api/process.html#process_process_send_message_sendhandle_options_callback

src/vpn-api/worker.ts
wrboyce: yes, the actual format is `>BYTECOUNT_CLI:` but the `>BYTECOUNT` part is trimmed from the library (which is admittedly a bit hacky, the library doesn't seem to be expecting CLI bytecount messages, just `>BYTECOUNT` messages which are global counters, but it works -- I guess ideally we would extend the openvpn-telnet library to know about bytecount_cli messages).

src/vpn-api/worker.ts
wrboyce: There we go, that explains it. I've removed the check on `process` though, that wasn't necessary.

src/vpn-api/worker.ts
wrboyce: I've PR'd `telnet-openvpn`, if it doesn't get a response soon-ish I'll just fork the repo and publish under `@balena`

wrboyce: Nice, he’s merged it and released a new version. I’ll get this branch updated.

src/vpn-api/worker.ts
```diff
@@ -143,13 +153,68 @@ const worker = (instanceId: number) => {
...
+					if (data.bytecount_cli == null) {
+						return;
+					}
+					const clientId = data.bytecount_cli[0];
+					if (clientMap[clientId] == null) {
+						logger.error(`Unknown CID(${clientId}) from OpenVPN!`);
+						return;
+					}
+
+					const uuid = clientMap[clientId];
```
Page-: Can you move this up as
```ts
const uuid = clientMap[clientId];
if (uuid == null) {

File: src/vpn-api/app.ts [hunk 9]
```diff
@@ -35,34 +39,103 @@ import worker from './worker';
...
+	} = {};
+
+	if (VPN_BYTECOUNT_INTERVAL > 0) {
+		cluster.on('message', (_worker, msg) => {
+			const { uuid, ts, rxBytes, txBytes } = msg;
+			sessions[uuid] = { ts, rxBytes, txBytes };
 		});
 	}
+
+	const kb = 2 ** 10;
```
Page-: I know this is correct but 1024 would be clearer to me - I don't know for other people though

src/vpn-api/worker.ts
```diff
@@ -143,13 +153,68 @@ const worker = (instanceId: number) => {
...
+				vpn.connection.shellPrompt = '';
+				// monitor new client connections and map cid to uuid
+				vpn.on('log', data => {
+					if (data.includes('CLIENT:ESTABLISHED,')) {
+						const clientId = data.split(',')[1].trim();
+						const idMapper = (logData: string) => {
+							if (logData.includes('CLIENT:ENV,common_name=')) {
+								const uuid = logData.split('=')[1].trim();
+								// expire any previous mappings for this device
+								_.forOwn(clientMap, (client, key) => {
```
Page-: I've got a feeling this could be a performance issue in prod, I think it might be worth having two maps one clientId -> uuid and one uuid -> clientId so that you can change this to
```ts
delete clientMap[uuidMap[uuid]]
```
so that it becomes a constant time and also avoid the potentially dangerous delete on an object you're iterating (no idea if it is dangerous, but it's definitely something that often is).

src/vpn-api/worker.ts
wrboyce: Done.

src/vpn-api/worker.ts
wrboyce: Done, I must admit, I had similar concerns.

src/vpn-api/app.ts
wrboyce: Added a comment.

config/confd/conf.d/haproxy.toml
```diff
@@ -1,4 +1,7 @@
 [template]
 src = "haproxy.cfg.tmpl"
 dest = "/etc/haproxy/haproxy.cfg"
-keys = [ "/resin/vpn/" ]
+keys = [
+    "/resin/monitor/secret-token",
```
dt-rush: This is now `/balena/monitor/secret-token`

config/confd/templates/haproxy.cfg.tmpl
```diff
@@ -29,21 +28,38 @@ defaults
 	timeout client {{$timeout}}s
 	timeout server {{$timeout}}s
 
+{{ if ne (getv "/resin/monitor/secret-token" "") "" -}}
```
dt-rush: (this is therefore renamed as well)

config/confd/templates/haproxy.cfg.tmpl
```diff
@@ -29,21 +28,38 @@ defaults
...
+userlist metrics
+	user monitor insecure-password "{{ getv "/resin/monitor/secret-token" }}"
```
dt-rush: (this is therefore renamed as well)

config/confd/templates/haproxy.cfg.tmpl
```diff
@@ -29,21 +28,38 @@ defaults
...
 frontend health-80
 	mode http
 	option httplog
 	bind ipv4@:80
 	bind ipv4@:81 accept-proxy
-	acl vpn_dead nbsrv(vpn-cluster) lt 1
+	acl vpn_dead nbsrv(vpn-workers) lt 1
 	monitor-uri /ping
 	monitor fail if vpn_dead
+{{- if ne (getv "/resin/monitor/secret-token" "") "" -}}
```
dt-rush: (this is therefore renamed as well)

src/vpn-api/app.ts
```diff
@@ -35,34 +39,103 @@ import worker from './worker';
...
+		tb,
+	];
+
+	// setup metrics for prometheus
+	metrics.describe(
+		'vpn_sessions_rx_bytes',
+		'histogram of rx bytes per vpn session',
+		{
+			buckets,
+			labelNames: [],
```
dt-rush: You can just omit `labelNames` in this case

src/vpn-api/app.ts
```diff
@@ -35,34 +39,103 @@ import worker from './worker';
...
+		},
+	);
+	metrics.describe(
+		'vpn_sessions_tx_bytes',
+		'histogram of tx bytes per vpn session',
+		{
+			buckets,
+			labelNames: [],
```
dt-rush: ditto here

src/vpn-api/app.ts
```diff
@@ -35,34 +39,103 @@ import worker from './worker';
...
+		},
+	);
+
+	const app = express();
+	app.disable('x-powered-by');
+	app.get('/ping', (_req, res) => res.send('OK'));
+	app.use(morgan('combined'));
+	app.get('/metrics', (_req, res) => {
```
dt-rush: There's actually, in `@balena/node-metrics-gatherer` @ `v2.2.1`, a better pattern for aggregating metrics of `cluster` workers, which uses a pattern designed by the prometheus client library for this purpose. See [the changed `index.js` in `resin-api` for an example](https://github.com/balena-io/resin-api/blob/b95ad206402a009c80b3902f829ec3dcc9d19fc0/index.js).

In this case, you'd do something like:

- replace `process.send` in the `cluster` worker code with a simple call to metrics, like `metrics.histogram('vpn_session_rx_bytes', rxBytes)`

- in the master thread, do:

```
app.get('/cluster_metrics', metrics.aggregateRequestHandler());
``` 

(replaceing the entire block from L125 to L134)

dt-rush: That is to say, the workers perform metrics calls as if they weren't in a cluster, while the master, in the internals of `prom-client` which `metrics.aggregateRequestHandler()` wraps, will manage the IPC and summation logic which gathers each cluster workers' metrics when the time comes to export.

src/vpn-api/worker.ts
```diff
@@ -143,13 +155,69 @@ const worker = (instanceId: number) => {
...
+						logger.error(`Unknown CID(${clientId}) from OpenVPN!`);
+						return;
+					}
+
+					const ts = Date.now();
+					const rxBytes = parseInt(data.bytecount_cli[1], 10);
+					const txBytes = parseInt(data.bytecount_cli[2], 10);
+					// send bytecount data to master process
+					if (process.send != null) {
+						process.send({ uuid, ts, rxBytes, txBytes });
```
dt-rush: This would be replaced with:

```
metrics.histogram('vpn_session_rx_bytes', rxBytes);
metrics.histogram('vpn_session_tx_bytes', txBytes);
```

dt-rush: (thus obviating the call to `Date.now()` and the check for `process.send != null`)

src/vpn-api/app.ts
wrboyce: Oh sweet, thanks for the heads up!

src/vpn-api/worker.ts
```diff
@@ -143,13 +187,66 @@ const worker = (instanceId: number) => {
 			),
 		)
 		.tap(() => openvpn.start())
+		.delay(1000)
```
Page-: It would be nice to not have to rely on an arbitrary delay if possible

wrboyce:
Superseded by #42

PR #42, opened 2019-04-10
vpn: use management interface to manage openvpn processes

wrboyce:
* Handle client authentication via management events.
* Generate custom, more fine grained, logs; openvpn log verbosity can be
  toggled with SIGUSR2.
* Use `child_process.spawn` in place of `forever.Monitor` meaning a process
  exit will cause the worker process it be restarted meaning the restart will
  pass through the initialisation flow again.
* Gather client events and publish for Prometheus (session bitrate + length,
  online clients, total clients, auth failures).

Change-type: minor

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---


src/vpn-api/utils/openvpn.ts
```diff
@@ -0,0 +1,287 @@
...
+					this.emit(`client:${eventName}`, ...args);
+				}
+			}
+		};
+		return emitter;
+	};
+
+	public start() {
+		return Promise.try(() => this.process.start())
+			.delay(1000)
```
wrboyce: As discussed, Page, I'll come up with something nicer here before we merge.

src/vpn-api/app.ts
```diff
@@ -16,53 +16,71 @@
...
+
+	'VPN_BASE_SUBNET',
+	'VPN_BASE_PORT',
+	'VPN_BASE_MANAGEMENT_PORT',
+	'VPN_INSTANCE_SUBNET_BITMASK',
+]
 	.filter(key => process.env[key] == null)
 	.forEach((key, idx, keys) => {
-		logger.error(`${key} env variable is not set.`);
+		logger.emerg(`${key} env variable is not set.`);
```
hedss: What's the difference between a `crit` and an `emerg`?

src/vpn-api/utils/haproxy.ts
```diff
@@ -0,0 +1,47 @@
...
+	This program is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import * as Promise from 'bluebird';
```
hedss: I think most other places, we're switching to `import * as Bluebird from 'bluebird'`.

src/vpn-api/utils/openvpn.ts
```diff
@@ -0,0 +1,308 @@
...
+	This program is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import * as Promise from 'bluebird';
```
hedss: `Bluebird` instead of `Promise`.

src/vpn-api/worker.ts
```diff
@@ -71,91 +46,191 @@ const getInstanceSubnet = (instanceId: number) => {
...
 
 const worker = (instanceId: number) => {
-	logger.info(`worker-${instanceId} process started with pid ${process.pid}`);
+	const logger = getLogger('vpn', instanceId);
+
+	logger.notice(`process started with pid=${process.pid}`);
+
+	const fatalErrorHandler = (err: Error) => {
+		logger.emerg(`${err.message}`);
+		process.exit(2);
```
hedss: Worth using `process.exitCode` instead, here? And in fact, maybe everywhere else where it's called.

src/vpn-api/worker.ts
```diff
@@ -71,91 +46,191 @@ const getInstanceSubnet = (instanceId: number) => {
...
+		metrics.gauge('vpn_online_devices', vpn.nClients);
+		if (data.bytes_received != null && data.bytes_sent != null) {
+			metrics.histogram('vpn_session_rx_bytes', data.bytes_received);
+			metrics.histogram('vpn_session_tx_bytes', data.bytes_sent);
+		}
+		return clients.disconnected(data).then(logStateUpdate);
+	});
+
+	// setup metrics for prometheus
+	const kb = 2 ** 10; // 1024
```
hedss: @Page- is going to ask for these to be explicit instead of powers of. :D

config/confd_env_backend/templates/env.tmpl
```diff
@@ -6,6 +6,7 @@ SENTRY_DSN={{getenv "SENTRY_DSN"}}
 BALENA_API_HOST={{getenv "BALENA_API_HOST"}}
 
 IPTABLES_FORWARD_POLICY=DROP
+VPN_BYTECOUNT_INTERVAL=10
 BALENA_VPN_GATEWAY={{getenv "BALENA_VPN_GATEWAY"}}
 VPN_INSTANCE_COUNT={{ if eq (getenv "PRODUCTION_MODE") "true" }}{{getenv "NPROC"}}{{ else }}1{{ end }}
```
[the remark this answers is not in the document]
wrboyce: Nice catch, was testing I hadn't broken anything and accidentally included it in a commit.

src/vpn-api/utils/haproxy.ts
wrboyce: Interesting, okay. Do you know why?

src/vpn-api/worker.ts
wrboyce: Haha, the comment is there as an accepted compromise after a review from Page (in an earlier, ummerged, PR which this is superseding).

config/confd_env_backend/templates/env.tmpl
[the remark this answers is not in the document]
wrboyce: Wait, sorry -- misread your comment and thought I'd accidentally included my testing change which _did_ change it from `1` to `2`.

I'm not sure I follow what you mean, you could elaborate?

src/vpn-api/utils/haproxy.ts
hedss: I believe it's in a move to slowly move to native Promises where possible, only using Bluebird as sugar.

src/vpn-api/worker.ts
hedss: Fair enough, I retract the comment. ;-)

src/vpn-api/utils/haproxy.ts
wrboyce: Fair enough, I can certainly see the argument for using native Promises but Bluebird's sugar certainly does… _make the medicine go down_.

I'll get my coat.

config/confd_env_backend/templates/env.tmpl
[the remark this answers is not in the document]
wrboyce: 🤦‍♂️

config/confd_env_backend/conf.d/haproxy.toml
```diff
@@ -3,6 +3,7 @@ src = "haproxy.cfg.tmpl"
 dest = "/etc/haproxy/haproxy.cfg"
 keys = [
   "BALENA_VPN_PORT",
+  "METRICS_SECRET_TOKEN",
```
hedss: `MONITOR_SECRET_TOKEN`

src/vpn-api/worker.ts
```diff
@@ -71,91 +46,194 @@ const getInstanceSubnet = (instanceId: number) => {
...
+	vpn.on('client:established', (clientId, data) => {
+		logger.info(
+			`connection established with client_id=${clientId} uuid=${data.username}`,
+		);
+		metrics.inc('vpn_online_devices');
+		return clients.connected(data).then(logStateUpdate);
+	});
+
+	vpn.on('client:bytecount', (_clientId, data) => {
+		metrics.histogram('vpn_session_rx_bytes', data.bytes_received);
```
dt-rush: this should be `_sessions_`

src/vpn-api/worker.ts
```diff
@@ -71,91 +46,194 @@ const getInstanceSubnet = (instanceId: number) => {
...
+		metrics.histogram('vpn_session_tx_bytes', data.bytes_sent);
```
dt-rush: this should be `_sessions_`

src/vpn-api/app.ts
```diff
@@ -16,53 +16,71 @@
...
+	logger.notice(
+		`open-balena-vpn@${VERSION} process started with pid=${process.pid}`,
+	);
+	spawnChildren(VPN_INSTANCE_COUNT, logger);
+
+	const app = express();
+	app.disable('x-powered-by');
+	app.get('/ping', (_req, res) => res.send('OK'));
+	app.use(morgan('combined'));
+	app.get('/metrics', metrics.aggregateRequestHandler());
```
dt-rush: should be `/cluster_metrics`

wrboyce:
**Note to reviewers**: Github has decided that I wrote too much code in `src/vpn-api/worker.ts` and is hiding the diff. It is _very_ easy to overlook this as "file deleted" or "oh thats just package-lock.json", so watch out :)

src/vpn-api/app.ts
```diff
@@ -19,50 +19,58 @@ import * as cluster from 'cluster';
...
+		logger.emerg(`${key} env variable is not set.`);
 		if (idx === keys.length - 1) {
-			process.exit(1);
+			process.exitCode = 1;
 		}
 	});
 
 const VPN_INSTANCE_COUNT =
 	parseInt(process.env.VPN_INSTANCE_COUNT!, 10) || os.cpus().length;
+const VPN_VERBOSE_LOGS = process.env.VPN_VERBOSE_LOGS! === 'true';
```
Page-: Is this `!` actually necessary?

src/utils/index.ts
```diff
@@ -29,11 +33,51 @@ export const balenaApi = new PinejsClientRequest(
...
+		const workerId = i + 1;
+		const restartWorker = (code?: number, signal?: string) => {
+			if (signal != null) {
+				logger.crit(`worker-${workerId} killed with signal ${signal}`);
+			}
+			if (code != null) {
+				logger.crit(`worker-${workerId} exited with code ${code}`);
+			}
+			const env = {};
+			_.merge(env, process.env, { WORKER_ID: workerId });
```
Page-: I'm assuming you actually want `_.assign` behaviour here rather than `_.merge`, in which case you can do
```suggestion
			const env = {
				...process.env,
				WORKER_ID: workerId
			};
```

src/connect-proxy/worker.ts
```diff
@@ -133,17 +132,21 @@ class Tunnel extends nodeTunnel.Tunnel {
...
-		return Promise.try(() => parseRequest(req)).then(({ uuid, auth }) =>
+		return Bluebird.try(() => parseRequest(req)).then(({ uuid, auth }) =>
 			lookupAsync(`${uuid}.vpn`)
 				.then(() => {
 					logger.info(`connecting to ${host}:${port}`);
-					return super.connect(port, host, client, req);
+					return super.connect(port, host, client, req).tap(socket => {
+						metrics.inc(Metrics.ActiveTunnels);
+						metrics.inc(Metrics.TotalTunnels);
+						socket.on('close', () => metrics.dec(Metrics.ActiveTunnels));
```
Page-: Probably
```suggestion
						socket.on('close', () => { metrics.dec(Metrics.ActiveTunnels) });
```

test/app.ts
```diff
@@ -55,16 +55,16 @@ const vpnDefaultOpts = [
...
+	closeAsync(): Bluebird<HttpServerAsync>;
 }
 
 before(() => {
 	chai.use(chaiAsPromised);
 });
 
 describe('vpn worker', function() {
-	this.timeout(10 * 1000);
+	this.timeout(15 * 1000);
```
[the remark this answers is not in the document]
wrboyce: Connecting to management port, etc.

test/app.ts
```diff
@@ -34,7 +34,7 @@ import vpnWorker from '../src/vpn-api/worker';
 const vpnHost = process.env.VPN_HOST || '127.0.0.1';
 const vpnPort = process.env.VPN_PORT || '443';
 const caCertPath =
-	process.env.CA_CERT_PATH || path.resolve(__dirname, '../openvpn/ca.crt');
+	process.env.CA_CERT_PATH || path.resolve('/etc/openvpn/ca.crt');
```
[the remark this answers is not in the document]
wrboyce: D'oh, of course. Thanks.

src/vpn-api/worker.ts
```diff
@@ -71,91 +71,245 @@ const getInstanceSubnet = (instanceId: number) => {
...
+			msg = `process terminated with signal ${signal}`;
+		}
+		fatalErrorHandler(new Error(msg));
+	});
+	vpn.on('process:error', fatalErrorHandler);
+
+	// capture openvpn logs
+	vpn.on('log', (level, message) => {
+		// only log warnings (or more severe) unless verbose=true
+		if (verbose || logger.levels[level] <= logger.levels.warning) {
```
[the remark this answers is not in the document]
wrboyce: By setting the transport log level in Winston? This line is only filtering excessively verbose messages from openvpn, there are still notice/info/debug level messages from the application that we want to display (even in non-verbose mode).

src/vpn-api/worker.ts
```diff
@@ -71,91 +71,245 @@ const getInstanceSubnet = (instanceId: number) => {
...
 
 const worker = (instanceId: number) => {
-	logger.info(`worker-${instanceId} process started with pid ${process.pid}`);
+	const logger = getLogger('vpn', instanceId);
+	const clientLog: { [key: string]: number } = {};
+
+	logger.notice(`process started with pid=${process.pid}`);
+
+	const fatalErrorHandler = (err: Error) => {
+		logger.emerg(err.message);
```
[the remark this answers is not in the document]
wrboyce: `fatal` is not a valid RFC 5424 severity level.

src/vpn-api/utils/openvpn.ts
```diff
@@ -0,0 +1,308 @@
...
+					for (const arg of eventArgs) {
+						args.push(parsePossibleInt(arg));
+					}
+				}
+				args.push(env);
+				this.off('manager:data', emitter);
+				this.on('manager:data', this.dataHandler);
+				this.emit(`client:${eventName}`, ...args);
+				// re-emit the last message if it was not `CLIENT:ENV,END`
+				if (data.substr(1) !== 'CLIENT:ENV,END') {
```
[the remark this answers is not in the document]
wrboyce: It is a `>`, I felt it made everything a little less readable to have `>` everywhere in the strings.

src/vpn-api/utils/openvpn.ts
```diff
@@ -0,0 +1,308 @@
...
+	private clientEventEmitterFactory = (
+		eventName: string,
+		clientId: number,
+		eventArgs?: string[],
+	) => {
+		const env: { [key: string]: number | string } = {};
+		const emitter = (data: string) => {
+			if (data.substr(1, 10) === 'CLIENT:ENV' && data.substr(12) !== 'END') {
+				// >CLIENT:ENV,key=val
+				const msg = data.substr(12);
```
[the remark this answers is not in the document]
wrboyce: You mean like `const msg = data.substr('>CLIENT:ENV'.length);`?

src/vpn-api/utils/openvpn.ts
```diff
@@ -0,0 +1,308 @@
...
+		private mgtPort: number,
+		private subnet: Netmask,
+		private gateway?: string,
+	) {
+		super();
+
+		this.connector = new VpnConnector();
+		// proxy `data` events from connector, splitting at newlines
+		this.connector.connection.on('data', data => {
+			const lines = ((this.buf || '') + data.toString()).split('\r\n');
```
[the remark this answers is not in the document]
wrboyce: `\r\n` afaik but it wouldn't hurt to split on `/\r?\n/`.

src/vpn-api/utils/clients.ts
```diff
@@ -78,59 +79,50 @@ const setDeviceState = (() => {
...
 			deviceStates[uuid] = {
 				targetState: {},
 				currentState: {},
-				promise: Promise.resolve(),
+				promise: Bluebird.resolve(),
 			};
 		}
 		deviceStates[uuid].targetState = state;
-		applyState(uuid);
+		return applyState(uuid);
```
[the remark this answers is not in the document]
wrboyce: So I can handle the loggging in `worker.ts` (`clients.connect(data).then(logStateUpdate)`).

src/vpn-api/worker.ts
```diff
@@ -71,91 +71,245 @@ const getInstanceSubnet = (instanceId: number) => {
...
+					vpnPort,
+				),
+			)
+			.tap(() =>
+				logger.info(
+					`registered as haproxy backend server vpn-workers/vpn${instanceId}`,
+				),
+			)
+			.catch(fatalErrorHandler)
+			.tap(() => logger.notice(`waiting for clients...`))
```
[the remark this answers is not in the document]
wrboyce: You've flagged this pattern a few time (`.tap(() => blah())`) showing a preference for `.tap(() => {blah()})`.

Is this just a preference or is there a technical benefit?

src/vpn-api/worker.ts
```diff
@@ -71,91 +71,245 @@ const getInstanceSubnet = (instanceId: number) => {
...
+				data.bytes_sent
+			})`;
+			metrics.histogram(Metrics.SessionRxBytes, data.bytes_received);
+			metrics.histogram(Metrics.SessionTxBytes, data.bytes_sent);
+			writeBitrateMetrics(data as VpnClientBytecountData, data.time_duration);
+		}
+		logger.info(msg);
+		metrics.dec(Metrics.OnlineDevices);
+		delete clientLog[clientId];
+		return clients.disconnected(data).then(logStateUpdate);
```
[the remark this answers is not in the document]
wrboyce: I don't think it can result in a rejection… If you take a look at the code in `clients.ts` it has a catch on itself which calls itself after a delay.

As you noted later in the review, I'm now returning the final promise from the client.{connect, disconnect} functions. Prior to this PR, the returned promise from these calls was being ignored anyway.

src/vpn-api/utils/openvpn.ts
```diff
@@ -0,0 +1,308 @@
...
+			this.emit('process:exit', code, signal);
+		});
+
+		return this.waitForStart();
+	}
+
+	private waitForStart(since?: number): Bluebird<true> {
+		const now = Date.now();
+		return new Bluebird((resolve, reject) => {
+			if (since != null && now - since > 5000) {
```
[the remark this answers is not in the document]
wrboyce: Is the logic the same here? I'm not executing the promise once and waiting 5 seconds, I am repeatedly iterating over it until either it is resolved or 5 seconds have elapsed. See the call to `this.waitForStart` in the socket error handler. This line is to ensure that if after 5 seconds the socket is still unavailable we stop retrying.

src/vpn-api/worker.ts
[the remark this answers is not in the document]
Page-: It makes it clear to the reader that no value is ever returned which is the reason for the preference, but it also does the same for v8 which may allow it to make optimisations that it otherwise couldn't (whether or not it actually uses that info is another matter, and whether in some of these cases it can infer it regardless of the less clear pattern I cannot say though - but providing the info gives it the best chance).  Also in some cases returning a value that we don't care about may still end up in it getting used/stored, eg returning an object to bluebird will require it to do additional checks to see if it's a promise that it should be waiting on and if it was a `.then` instead of a `.tap` it has to store that value and pass it on as it doesn't know that it's actually unused.  Generally it's good practice imo to not return anything if you don't want it to be acted upon rather than relying on the current implementation details meaning that it won't be as those could change.

That said regardless of the technical reasons I'd still want this pattern because of the clarity to people reading the code rather than potentially having to double check that the return value actually isn't important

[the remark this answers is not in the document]
wrboyce: Thanks!

src/vpn-api/worker.ts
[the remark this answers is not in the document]
Page-: Ah, so there's a different level for the application vs openvpn?  It still might be worth looking into having winston separate the log levels for the two but this `if` makes sense now.  As an additional thing though it might be worth having `verbose` be a scale rather than a boolean, ie `logger.levels[level] <= verbosity` - but that can wait for a separate PR if at all

src/vpn-api/utils/openvpn.ts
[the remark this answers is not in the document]
Page-: The logic should be the same, the only thing that is wonky is that you'd add a timeout at each level but only the outermost level could ever actually time out, so ideally you'd have the recursion separate to the timeout so the timeout is only added once, eg
```ts
this.waitForStart().timeout(5000).catch(TimeoutError, () => {
	throw new Error('startup timeout')
})
```

src/vpn-api/utils/openvpn.ts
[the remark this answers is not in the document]
Page-: I think it'd be better in general to be able to avoid the need for a lot of the `substr` calls, since I expect these can be called a lot (that said v8 might be able to optimise them out but I wouldn't want to have to rely on it for what could be a fairly hot path)

src/vpn-api/utils/openvpn.ts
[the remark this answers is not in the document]
Page-: Yeah, that way it avoids the potential for things to fall out of sync.  Also because it's a common pattern here it might be worth adding a helper along the lines of
```ts
const x = (data: string, msg: string) => {
	if (data.startsWith(msg)) {
		return data.substr(msg.length);
	}
	return false
}
```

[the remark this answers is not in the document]
Page-: The only issue is because of needing to keep the result but also check it then the `if` statements can get a little awkward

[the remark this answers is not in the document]
wrboyce: Yeah I was just about to ask about that as your second comment landed!

src/vpn-api/utils/clients.ts
```diff
@@ -78,59 +79,50 @@ const setDeviceState = (() => {
...
-					let stateMsg = `common_name=${targetState.common_name} connected=${
-						targetState.connected
-					}`;
-					if (targetState.virtual_address != null) {
-						stateMsg = `${stateMsg} virtual_address=${
-							targetState.virtual_address
-						}`;
-					}
-					logger.info(`Successfully updated state for ${uuid}: ${stateMsg}`);
+					return targetState as DeviceState;
```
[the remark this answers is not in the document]
wrboyce: The `Partial` is required for when tracking the state of a previously unknown devices, we initialise the states as `{}`.

src/vpn-api/utils/openvpn.ts
```diff
@@ -0,0 +1,308 @@
...
+				return reject(new Error('startup timeout'));
+			}
+			const socket = new net.Socket();
+			const errorHandler = () => {
+				socket.destroy();
+				resolve(this.waitForStart(now));
```
Page-: I think this would be better as
```ts
reject(new TimeoutError('startup timeout'))
...
reject(new Error('..'))
...
.catch((err) => {
	if (err instanceof TimeoutError) {
		throw err
	}
	return this.waitForStart(now))
}
```
to get back into a standard promise chain as soon as possible rather than having the current unusual pattern

src/vpn-api/utils/clients.ts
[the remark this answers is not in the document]
Page-: You mean the
```ts
		if (deviceStates[uuid] == null) {
			deviceStates[uuid] = {
				targetState: {},
				currentState: {},
				promise: Bluebird.resolve(),
			};
		}
		deviceStates[uuid].targetState = state;
```
bit below?  In that case it can be:
```ts
		if (deviceStates[uuid] == null) {
			deviceStates[uuid] = {
				targetState: state,
				currentState: {},
				promise: Bluebird.resolve(),
			};
		} else {
			deviceStates[uuid].targetState = state;
		}
```
in order to avoid the `Partial` on `targetState`

src/vpn-api/utils/openvpn.ts
```diff
@@ -185,8 +185,8 @@ export class VpnManager extends EventEmitter {
 				'established',
 				'disconnect',
 			]) {
-				if (data.substr(0, eventType.length) === eventType.toUpperCase()) {
-					const eventData = data.substr(8);
+				const eventData = data.substr('>CLIENT:'.length);
```
Page-: This should be outside of the `for` loop so that it only needs to be done once

src/vpn-api/utils/clients.ts
```diff
@@ -78,59 +79,56 @@ const setDeviceState = (() => {
...
+		return applyState(uuid);
 	};
 })();
 
-export const connected = (data: Partial<DeviceState>) => {
-	data = _.pick(data, 'common_name', 'virtual_address');
-	data.connected = true;
-	return setDeviceState(data as DeviceState);
+export const connected = (data: VpnClientTrustedData) => {
+	const state = {
```
Page-: I think this is better as
```suggestion
	const state: DeviceState = {
```
and removing the cast below

src/vpn-api/utils/clients.ts
```diff
@@ -78,59 +79,56 @@ const setDeviceState = (() => {
...
+	};
+	return setDeviceState(state as DeviceState);
 };
 
-export const disconnected = (data: Partial<DeviceState>) => {
-	data = _.pick(data, 'common_name');
-	data.connected = false;
-	return setDeviceState(data as DeviceState);
+export const disconnected = (data: VpnClientDisconnectData) => {
+	const state = {
```
Page-: I think this is better as
```suggestion
	const state: DeviceState = {
```
and removing the cast below

src/vpn-api/utils/openvpn.ts
wrboyce: Typo, fixed :)

src/vpn-api/worker.ts
```diff
@@ -71,91 +76,256 @@ const getInstanceSubnet = (instanceId: number) => {
...
+			});
+	});
+
+	vpn.on('client:established', (clientId, data) => {
+		logger.info(
+			`connection established with client_id=${clientId} uuid=${data.username}`,
+		);
+		metrics.inc(Metrics.OnlineDevices);
+		metrics.inc(Metrics.TotalDevices);
+		[clientLog[clientId]] = process.hrtime();
```
Page-: I had no idea this was a thing

src/vpn-api/utils/openvpn.ts
```diff
@@ -0,0 +1,318 @@
...
+	}
+
+	public connect(): Bluebird<true> {
+		return Bluebird.try(() =>
+			this.connector
+				.connect({
+					port: this.mgtPort,
+					shellPrompt: '',
+				})
+				.then(() => this.emit('manager:connect')),
```
Page-:
```suggestion
		).then(() => { this.emit('manager:connect') }),
```

src/vpn-api/utils/openvpn.ts
```diff
@@ -0,0 +1,318 @@
...
+		private instanceId: number,
+		private vpnPort: number,
+		private mgtPort: number,
+		private subnet: Netmask,
+		private gateway?: string,
+	) {
+		super();
+		// proxy `data` events from connector, splitting at newlines
+		this.connector.connection.on('data', data => {
+			const lines = ((this.buf || '') + data.toString()).split(/\r?\n/);
```
Page-: If you start with
```
private buf = ''
```
then you can make this just
```suggestion
			const lines = (this.buf + data.toString()).split(/\r?\n/);
```

src/vpn-api/utils/openvpn.ts
```diff
@@ -0,0 +1,318 @@
...
+export interface VpnClientDisconnectData
+	extends VpnClientUntrustedData,
+		Partial<VpnClientTrustedData>,
+		Partial<VpnClientBytecountData> {
+	time_duration: number;
+}
+
+export const isTrusted = (data: any): data is VpnClientTrustedData =>
+	data.common_name != null;
+export const hasBytecountData = (data: any): data is VpnClientBytecountData =>
```
Page-: It'd be better to have these as
```suggestion
export const hasBytecountData = (data: {}): data is VpnClientBytecountData =>
```
or similar to guarantee that it's an object passed in since the code assumes it'll be an object

src/vpn-api/utils/openvpn.ts
```diff
@@ -0,0 +1,318 @@
...
+		VpnClientTrustedData {}
+// CLIENT:DISCONNECT
+export interface VpnClientDisconnectData
+	extends VpnClientUntrustedData,
+		Partial<VpnClientTrustedData>,
+		Partial<VpnClientBytecountData> {
+	time_duration: number;
+}
+
+export const isTrusted = (data: any): data is VpnClientTrustedData =>
```
Page-: It'd be better to have these as
```suggestion
export const isTrusted = (data: {}): data is VpnClientTrustedData =>
```
or similar to guarantee that it's an object passed in since the code assumes it'll be an object

src/vpn-api/utils/haproxy.ts
```diff
@@ -0,0 +1,47 @@
...
+		name: string,
+		port: number,
+		host: string = '127.0.0.1',
+	): Bluebird<boolean> {
+		const preamble = `set server ${name}`;
+		return this.connect()
+			.then(socket =>
+				socket.write(
+					`${preamble} addr ${host} port ${port}\r\n${preamble} state ready\r\n`,
+					() => socket.destroy(),
```
Page-:
```suggestion
					() => { socket.destroy() },
```
?

src/vpn-api/utils/haproxy.ts
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	public register(
+		name: string,
+		port: number,
+		host: string = '127.0.0.1',
+	): Bluebird<boolean> {
+		const preamble = `set server ${name}`;
+		return this.connect()
+			.then(socket =>
```
Page-:
```suggestion
			.then(socket => {
```

src/vpn-api/utils/clients.ts
```diff
@@ -52,13 +53,13 @@ export interface DeviceState {
...
-	const applyState = (uuid: string) => {
-		deviceStates[uuid].promise = deviceStates[uuid].promise.then(() => {
+	const applyState = (uuid: string) =>
+		(deviceStates[uuid].promise = deviceStates[uuid].promise.then(() => {
 			// Get the latest target state at the start of the request
 			const { targetState, currentState } = deviceStates[uuid];
 			if (_.isEqual(targetState, currentState)) {
 				// If the states match then we don't have to do anything
-				return;
+				return targetState as DeviceState;
```
Page-:
```suggestion
				return targetState;
```

src/vpn-api/app.ts
```diff
@@ -19,50 +19,58 @@ import * as cluster from 'cluster';
...
+		logger.emerg(`${key} env variable is not set.`);
 		if (idx === keys.length - 1) {
-			process.exit(1);
+			process.exitCode = 1;
 		}
 	});
 
 const VPN_INSTANCE_COUNT =
 	parseInt(process.env.VPN_INSTANCE_COUNT!, 10) || os.cpus().length;
+const VPN_VERBOSE_LOGS = process.env.VPN_VERBOSE_LOGS === 'true';
```
Page-: Probably no reason to assign to this const since it's meant to be toggleable (and I don't think this var is used outside of initialising the toggle)

src/vpn-api/utils/openvpn.ts
wrboyce: Unfortunately `Array<T>.pop` returns `T | undefined` so the `|| ''` remains necessary.

src/vpn-api/utils/openvpn.ts
wrboyce: `{}` causes a type error in the body of the function ("Error:(63, 7) TS2339: Property 'common_name' does not exist on type '{}'.").

PR #52, opened 2019-04-17
perms: Add tunnel-any check when tunnelling ports

richbayliss:
Add an additional check to the actions allowed when tunnelling
ports through the connect-proxy. This means that if the key used
is sufficiently privileged then any port can be tunnelled to a
device.

Change-type: patch
Signed-off-by: Rich Bayliss <rich@balena.io>


Page-:
Can we delay this until after we deploy the metrics please, it should be fine just want to keep separate bits out for safety until we've got master deployed

richbayliss:
Sure thing @Page- makes sense

wrboyce:
Ditto :)

Issue #50, opened 2019-04-17
`connect-proxy` can attempt to forward connections via itself


wrboyce:
If the API incorrectly reports that a device is connected to a service instance, it can result in a connect proxy repeatedly forwarding the connection via itself.

wrboyce:
While this has been mostly mitigated, there still remains the possibility that an instance will restart (thus registering as a new service instance and receiving a new ID) and begin routing connections via itself in a loop (as the API will report the previous ID but the same IP). One possible solution for this is adding an `Forwarded: by=<id>` or similar header to forwarded requests and using that for loop detection.

Issue #48, opened 2019-04-17
register as single service instance


wrboyce:
As opposed to each worker registering as an individual `service_instance`, the master process should handle service registration and then make the service ID available to workers.

Issue #47, opened 2019-04-17
merge systemd services


wrboyce:
The `vpn-api` and `connect-proxy` services should be merged into a single systemd service so that state can be more easily shared between workers.

PR #53, opened 2019-04-17
refactor: merge systemd services

wrboyce:
* start all services and workers from a unified `open-balena-vpn`
  systemd service and master process
* register as a single service instance in master process and make
  available to workers
* prevent forwarding tunnel requests via proxy workers on the same instance

Connects-to: #47
Connects-to: #48 
Connects-to: #50 
Change-type: major

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---


wrboyce:
Putting this up as a draft for now as while it works, there are probably some improvements that can be made. Comments encouraged :)

src/utils/clients.ts
```diff
@@ -63,12 +62,13 @@ const setDeviceState = (() => {
...
-				url: `https://${BALENA_API_HOST}/services/vpn/client-${eventType}`,
-				timeout: REQUEST_TIMEOUT,
-				form: _.extend({ service_id: service.getId() }, targetState),
-				headers: { Authorization: `Bearer ${apiKey}` },
-			})
+			return request
+				.post({
+					url: `https://${BALENA_API_HOST}/services/vpn/client-${eventType}`,
+					timeout: REQUEST_TIMEOUT,
+					form: _.extend({ service_id: serviceId }, targetState),
```
Page-: I think this can be
```suggestion
					form: {
						service_id: serviceId,
						...targetState,
					},
```

Page-: Why is this `form` and not `json`?

src/utils/clients.ts
```diff
@@ -63,12 +62,13 @@ const setDeviceState = (() => {
 			}
 
 			const eventType = targetState.connected ? 'connect' : 'disconnect';
-			return pooledPostAsync({
-				url: `https://${BALENA_API_HOST}/services/vpn/client-${eventType}`,
-				timeout: REQUEST_TIMEOUT,
-				form: _.extend({ service_id: service.getId() }, targetState),
-				headers: { Authorization: `Bearer ${apiKey}` },
-			})
+			return request
```
Page-: Why is this no longer pooled?

src/proxy-worker.ts
```diff
@@ -0,0 +1,286 @@
...
+			const earlyError = (err: Error) => {
+				let errMsg = 'could not connect to vpn tunnel';
+				if (err != null && err.message) {
+					errMsg += `: ${err.message}`;
+				}
+				captureException(err, errMsg);
+				reject(new errors.RemoteTunnellingError(errMsg));
+			};
+			const proxyData = (chunk: Buffer) => {
+				if (chunk != null) {
```
Page-: You can change this to
```ts
if (chunk == null) {
	return;
}
tunnelProxyResponse += chunk.toString();
```
because you know that the
```ts
if (!_.includes(tunnelProxyResponse, '\r\n\r\n')) {
```
below won't change if there's no chunk.

That said is it even possible for the chunk to be null?

src/proxy-worker.ts
```diff
@@ -0,0 +1,286 @@
...
+						.tap(() => {
+							if (!data.is_connected_to_vpn) {
+								cltSocket.end('HTTP/1.0 503 Service Unavailable\r\n\r\n');
+								throw new errors.HandledTunnelingError(
+									`device not available: ${uuid}`,
+								);
+							}
+						}),
+				)
+				.tap(() => (req.url = `${uuid}.vpn:${port}`));
```
Page-:
```suggestion
				.tap(() => {req.url = `${uuid}.vpn:${port}`});
```

src/proxy-worker.ts
```diff
@@ -0,0 +1,286 @@
...
+			// we need to use VPN_SERVICE_API_KEY here as this could be an unauthenticated request
+			return device
+				.getDeviceByUUID(uuid, VPN_SERVICE_API_KEY)
+				.tap(data => {
+					if (data == null) {
+						cltSocket.end('HTTP/1.0 404 Not Found\r\n\r\n');
+						throw new errors.HandledTunnelingError(`device not found: ${uuid}`);
+					}
+				})
+				.tap(data =>
```
Page-: You can remove this `.tap` and join onto the one above
```suggestion
```

src/proxy-worker.ts
```diff
@@ -0,0 +1,286 @@
...
+						})
+						.then(serviceInstance => {
+							if (serviceInstance.id === this.serviceId) {
+								throw new errors.HandledTunnelingError(
+									'device is not available on registered service instance',
+								);
+							}
+							return serviceInstance.ip_address;
+						})
+						.then(vpnHost => {
```
Page-: This can be combined into the `.then` above
```suggestion
							const vpnHost = serviceInstance.ip_address;
```

src/utils/clients.ts
wrboyce: It is, it is just exported as `request` from `utils/index.ts`:

```ts
export { pooledRequest as request } from './request';
```

src/utils/clients.ts
wrboyce: Good question, looks like it has always been that way. I'll update it.

src/proxy-worker.ts
wrboyce: > That said is it even possible for the chunk to be null?

Hmm, I think not.

src/app.ts
```diff
@@ -96,38 +100,82 @@ if (cluster.isMaster) {
...
+				cluster.fork(env).on('exit', restartWorker);
+			};
+			restartWorker();
+		});
+
+		const app = express();
+		app.disable('x-powered-by');
+		app.get('/ping', (_req, res) => res.send('OK'));
+		app
+			.get('/cluster_metrics', (_req, res) => {
```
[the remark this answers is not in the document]
wrboyce: wow yeah, that is some pretty awful formatting. Blame `prettier` though!

test/app.ts
```diff
@@ -83,7 +83,10 @@ describe('vpn worker', function() {
 		expect(
 			service
 				.register()
-				.then(() => vpnWorker(1))
+				.tap(s => {
```
Page-:
```suggestion
				.then(serviceInstance => {
					instance = s;
					return vpnWorker(1, serviceInstance.getId());
				})
```

src/utils/device.ts
```diff
@@ -83,11 +83,16 @@ export const canAccessDevice = (
 		)
 		.catchReturn(false);
 
+interface VpnHost {
+	id: string;
```
Page-: This will actually be returned as a number I'm pretty sure
```suggestion
	id: number;
```

src/proxy-worker.ts
```diff
@@ -0,0 +1,285 @@
...
+								throw new errors.HandledTunnelingError(
+									`device not available: ${uuid}`,
+								);
+							}
+						}),
+				)
+				.tap(() => (req.url = `${uuid}.vpn:${port}`));
+		})
+			.then(() => next())
+			.catch(errors.APIError, err => {
```
Page-: Assuming `tunnelToDevice` is called regularly then there will be a definite cost to this sugar, so it's probably better to do:
```suggestion
			.catch(err => {
				if (err instanceof errors.APIError) {
					this.logger.alert(`Invalid Response from API (${err.message})`);
					cltSocket.end('HTTP/1.0 500 Internal Server Error\r\n\r\n');
				} else if (...) {
					...
				}
			})
```

src/proxy-worker.ts
```diff
@@ -0,0 +1,285 @@
...
+						.tap(() => {
+							if (!data.is_connected_to_vpn) {
+								cltSocket.end('HTTP/1.0 503 Service Unavailable\r\n\r\n');
+								throw new errors.HandledTunnelingError(
+									`device not available: ${uuid}`,
+								);
+							}
+						}),
+				)
+				.tap(() => (req.url = `${uuid}.vpn:${port}`));
```
Page-:
```suggestion
				.then(() => {
					req.url = `${uuid}.vpn:${port}`;
					next();
				})
```

src/proxy-worker.ts
```diff
@@ -0,0 +1,285 @@
...
+							if (!isAllowed) {
+								cltSocket.end(
+									'HTTP/1.0 407 Proxy Authorization Required\r\n\r\n',
+								);
+								throw new errors.HandledTunnelingError(
+									`device not accessible: ${uuid}`,
+								);
+							}
+						})
+						.tap(() => {
```
Page-: There's no need for this separate `.tap`, it can be part of the previous one
```suggestion
```

src/proxy-worker.ts
```diff
@@ -0,0 +1,285 @@
...
+							metrics.inc(Metrics.TotalTunnels);
+							socket.on('close', () => {
+								metrics.dec(Metrics.ActiveTunnels);
+							});
+						});
+					} else {
+						// The lookup failed so we try to forward to the correct vpn instance instead
+						return device
+							.getDeviceVpnHost(uuid, auth)
+							.catch(errors.APIError, err => {
```
Page-: Again depending on how often this is called then this sugar imposes a definite cost

src/app.ts
```diff
@@ -97,38 +97,78 @@ if (cluster.isMaster) {
...
+					SERVICE_ID: serviceInstance.getId(),
+				};
+				cluster.fork(env).on('exit', restartWorker);
+			};
+			restartWorker();
+		});
+
+		const app = express();
+		app.disable('x-powered-by');
+		app.get('/ping', (_req, res) => res.send('OK'));
```
Page-: Pretty sure this is changed in master to be an `app.use`

src/proxy-worker.ts
wrboyce: Really? That sucks :(

src/utils/device.ts
wrboyce: Does that mean registering a service instance registration will return a number, not a string?

src/app.ts
wrboyce: Not sure what you mean here? The `vpnWorker` uses `apiServer` (via `app.use`) for its internal API, but this `/ping` relates to the `express` server for exposing prometheus metrics.

```ts
		const app = express();
		app.disable('x-powered-by');
		app.get('/ping', (_req, res) => res.send('OK'));
```

I could abstract these three lines out into a common function, as they are used by both `api.ts` and `app.ts`.

src/proxy-worker.ts
wrboyce: like this?

```ts
.catch(err => {
	if (err instanceof errors.APIError) {
		this.logger.crit(
			`error connecting to device ${uuid} on port ${port} (${err.message})`,
		);
		throw new errors.HandledTunnelingError(err.message);
	}
	throw err;
})
```

src/proxy-worker.ts
Page-: It's a shame but at the same time the new format is how you'd have to do it with await/async so it at least makes that later transition easier

src/app.ts
Page-: I think I was just getting confused with the proxy

Issue #51, opened 2019-04-17
client auth requests to api should use pooled request workers


wrboyce:
This will prevent overwhelming the API when we have an influx of client connections (such as following a deploy).

PR #59, opened 2019-04-24
vpn: use pooled request workers for client authentication via api

wrboyce:
Connects-to: #51
Change-type: patch

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---

### DIFF
--- a/src/vpn-api/utils/clients.ts
+++ b/src/vpn-api/utils/clients.ts
@@ -32,7 +32,7 @@ import * as _ from 'lodash';
 import { apiKey, captureException } from '../../utils';
 
 import { VpnClientTrustedData } from './openvpn';
-import { pooledPostAsync } from './request';
+import { pooledRequest } from './request';
 import { service } from './service';
 
 const BALENA_API_HOST = process.env.BALENA_API_HOST!;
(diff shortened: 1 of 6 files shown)


src/vpn-api/utils/index.ts
```diff
@@ -20,5 +20,5 @@ export { clients };
 export { HAProxy } from './haproxy';
 export { Netmask } from './netmask';
 export { VpnManager } from './openvpn';
-export { request } from './request';
+export { pooledRequest as request } from './request';
```
Page-: Can you avoid renaming this when re-exporting please, it means that further down we can't tell that it's pooled
```suggestion
export { pooledRequest } from './request';
```
alternatively downstream can just import directly from the `./request` module

src/vpn-api/utils/clients.ts
```diff
@@ -32,7 +32,7 @@ import * as _ from 'lodash';
 import { apiKey, captureException } from '../../utils';
 
 import { VpnClientTrustedData } from './openvpn';
-import { pooledPostAsync } from './request';
+import { pooledRequest as request } from './request';
```
Page-: I'd prefer this to be
```suggestion
import { pooledRequest } from './request';
```
so that the usage is clear that it's pooled

PR #60, opened 2019-04-24
replace deprecated raven package with @sentry/node

wrboyce:
Change-type: patch

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---

### DIFF
--- a/src/connect-proxy/worker.ts
+++ b/src/connect-proxy/worker.ts
@@ -115,13 +115,7 @@ const tunnelToDevice: nodeTunnel.Middleware = (req, cltSocket, _head, next) =>
 			cltSocket.end('HTTP/1.0 403 Forbidden\r\n\r\n'),
 		)
 		.catch((err: Error) => {
-			captureException(
-				err,
-				`unexpected error establishing tunnel to ${req.url} (${err.message})`,
-				{
-					req,
-				},
-			);
+			captureException(err, 'proxy-tunnel-error', { req });
 			cltSocket.end('HTTP/1.0 500 Internal Server Error\r\n\r\n');
 		});
 
@@ -207,7 +201,8 @@ const forwardRequest = (
 			if (err != null && err.message) {
 				errMsg += `: ${err.message}`;
 			}
-			captureException(err, errMsg);
+			logger.warning(errMsg);
+			captureException(err, 'proxy-forward-error');
 			reject(new errors.RemoteTunnellingError(errMsg));
 		};
 		const proxyData = (chunk: Buffer) => {
@@ -264,7 +259,7 @@ const worker = (port: string) => {
 			logger.crit(
 				`failed to connect to device (${err.message || err})\n${err.stack}`,
 			);
-			captureException(err);
+			captureException(err, 'proxy-connect-error');
 		}
 	});
 	return tunnel;
--- a/src/utils/errors.ts
+++ b/src/utils/errors.ts
@@ -15,30 +15,42 @@
 	along with this program.  If not, see <https://www.gnu.org/licenses/>.
 */
 
-import * as _Raven from 'raven';
+import * as Sentry from '@sentry/node';
 import { TypedError } from 'typed-error';
 
 import { VERSION } from '.';
 
-export const Raven = _Raven;
-
-Raven.config(process.env.SENTRY_DSN || false, {
-	captureUnhandledRejections: true,
+Sentry.init({
+	dsn: process.env.SENTRY_DSN,
+	debug: process.env.NODE_ENV === 'development',
 	release: VERSION,
 	environment: process.env.NODE_ENV,
-}).install();
+});
+export { Sentry };
 
 export const captureException = (
 	err: Error,
-	message?: string,
-	options?: _Raven.CaptureOptions,
-): string => {
-	options = options || {};
-	if (message) {
-		options.extra = options.extra || {};
-		options.extra.message = message;
-	}
-	return Raven.captureException(err, options);
+	fingerprint: string,
+	opts?: { tags?: { [key: string]: string }; req?: { [key: string]: any } },
+) => {
+	Sentry.configureScope(scope => {
+		scope.addEventProcessor(evt => {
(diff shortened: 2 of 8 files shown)


src/utils/errors.ts
```diff
@@ -15,31 +15,20 @@
...
 
 import { VERSION } from '.';
 
-export const Raven = _Raven;
-
-Raven.config(process.env.SENTRY_DSN || false, {
-	captureUnhandledRejections: true,
+Sentry.init({
+	dsn: process.env.SENTRY_DSN,
+	debug: process.env.NODE_ENV === 'dev',
```
Page-: Usually this would be either empty or `development` for dev: http://expressjs.com/en/4x/api.html#app.set

src/vpn-api/api.ts
```diff
@@ -123,10 +123,11 @@ interface ExpressAsync extends express.Express {
 export const apiServer = (instanceId: number) => {
 	const app = Promise.promisifyAll(express()) as ExpressAsync;
 	app.disable('x-powered-by');
+	app.use(Sentry.Handlers.requestHandler());
```
Page-: We removed this when using raven because it enables/uses domains which causes a memory leak - did that get changed so it no longer uses domains?

wrboyce: It still uses `domain` but the implementation pretty much unrecognisable from `raven`. Is the use of `domain` going to be a problem full-stop?

src/utils/errors.ts
wrboyce: Ah, nice catch, ta.

wrboyce:
@Page- I don't think `raven-node` is maintained anymore, last release was 8 months ago (also `@sentry/node` is written in typescript, which feels like a win).

src/vpn-api/api.ts
wrboyce: Interesting that they opted to use `domain` in the newer module despite it being deprecated. I suppose it makes sense if they're doing some fancy error-juggling and there isn't a replacement api available in node yet.

Page-: Yeah, there isn't really an alternative to domains for what they're trying to do, but the memory leak is in domains themselves so any usage will be problematic for us

wrboyce: Gotcha, I'll nix that handler for now then!

src/connect-proxy/worker.ts
```diff
@@ -115,13 +115,7 @@ const tunnelToDevice: nodeTunnel.Middleware = (req, cltSocket, _head, next) =>
 			cltSocket.end('HTTP/1.0 403 Forbidden\r\n\r\n'),
 		)
 		.catch((err: Error) => {
-			captureException(
-				err,
-				`unexpected error establishing tunnel to ${req.url} (${err.message})`,
-				{
-					req,
```
Page-: You should add this manual `req` passing back so that the sentry error can have the context info available

src/vpn-api/utils/service.ts
```diff
@@ -70,9 +70,7 @@ class ServiceInstance {
 			})
 			.return(true)
 			.catch(err => {
-				captureException(err, 'Failed to send a heartbeat to the API', {
-					tags: { service_id: this.getId() },
```
Page-: Is having the service id available in the error not useful anymore?

src/vpn-api/utils/clients.ts
```diff
@@ -82,9 +82,7 @@ const setDeviceState = (() => {
 					return targetState;
 				})
 				.catch(err => {
-					captureException(err, 'Error updating state', {
-						user: { uuid },
```
Page-: Keeping the uuid that had an issue is probably helpful

src/vpn-api/api.ts
```diff
@@ -95,7 +95,7 @@ export const apiFactory = (instanceId: number) => {
 				}
 			})
 			.catch(err => {
-				captureException(err, 'Proxy Auth Error', { req });
+				captureException(err);
```
Page-: Can you add the req back please

src/vpn-api/utils/service.ts
```diff
@@ -25,6 +25,15 @@ class ServiceInstance {
 
 	constructor(private interval: number = 10 * 1000) {}
 
+	private captureException(err: Error) {
+		const tags: { [key: string]: string } = {};
+		try {
```
Page-: Do you need this try/finally?  It wasn't there before

src/vpn-api/utils/clients.ts
```diff
@@ -83,9 +83,7 @@ const setDeviceState = (() => {
...
-					});
+					captureException(err, { tags: { uuid } });
```
Page-: Didn't sentry treat the previous `user` property specially?  Is it worth keeping the new version using the `user` tracking stuff

src/vpn-api/utils/service.ts
```diff
@@ -70,9 +79,7 @@ class ServiceInstance {
 			})
 			.return(true)
 			.catch(err => {
-				captureException(err, 'Failed to send a heartbeat to the API', {
```
Page-: You've removed the message info from all the captures, I think it'd make sense to keep them so we get more info about where an error is occurring rather than just the stack trace (which depending on the error might be completely useless due to happening in an async callback)

src/vpn-api/utils/service.ts
wrboyce: Yes, if `getId` is called before the instance is registered it will thrown an error, and it possible we want to capture exceptions during registration.

src/vpn-api/utils/clients.ts
wrboyce: I'm not sure special sentry treats it versus other tags, but it does specifically limit the keys you can use to `id`, `username`, `email`, and `ip_address`.

src/vpn-api/utils/service.ts
wrboyce: Added fingerprints as discussed.

PR #61, opened 2019-04-25
dev: move typescript+@types dependencies to devDependencies

wrboyce:
Change-type: minor

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---


Dockerfile
```diff
@@ -21,18 +38,12 @@ RUN tmp="$(mktemp -d)" set -x \
 
 COPY package.json package-lock.json /usr/src/app/
 RUN npm ci --unsafe-perm --production && npm cache clean --force 2>/dev/null
-COPY . /usr/src/app
 
-COPY openvpn /etc/openvpn
+COPY src /usr/src/app/src
+COPY --from=builder /usr/src/app/build /usr/src/app/build
+COPY bin /usr/src/app/bin
```
Page-: I think you can condense these with
```suggestion
COPY src bin config openvpn /usr/src/app/
```

src/vpn-api/utils/index.ts
```diff
@@ -20,5 +20,5 @@ export { clients };
 export { HAProxy } from './haproxy';
 export { Netmask } from './netmask';
 export { VpnManager } from './openvpn';
-export { pooledRequest } from './request';
+export { request, pooledRequest } from './request';
```
Page-: ?

Dockerfile
wrboyce: Unfortunately not, that copies the contents of the directories.

src/vpn-api/utils/index.ts
wrboyce: It needs to be exported for the tests… I'm not entirely sure how the previous PR which changed the naming of this variable passed. Although now you point it out I'll just import `request` directly from `utils/request.ts` in the tests…

Dockerfile
Page-: Ah
```suggestion
COPY src/ bin/ config/ openvpn/ /usr/src/app/
```

wrboyce: Same result, afaik it isn't possible to copy multiple directories with a single `COPY`.

automation/test.sh
```diff
@@ -26,6 +26,7 @@ trap cleanup EXIT
 test_id=$(docker run --privileged -d \
 	--tmpfs /run \
 	--tmpfs /sys/fs/cgroup \
+	-v "${PWD}":/usr/src/app \
```
Page-: This looks like a leftover from local testing?

wrboyce: It was needed because of how the tests are run: We build the open-balena-vpn container, then run a custom set of commands inside it before invoke the tests. The problem being that the open-balena-vpn container no longer contains the files necessary to run the tests. As far as I could see this left me with two options. Either bind-mount the source folder as above, or make use of `docker cp` to copy the test files into the container.

Perhaps the latter, being more explicit, would be better?

Issue #67, opened 2019-06-22
Allow for customisation of vpn subnets


wrboyce:
See balena-io/balena#1837

PR #69, opened 2019-06-23
subnets: add config opts for vpn subnets and default to RFC6598 block

wrboyce:
Using CGN address space should eliminate all instances of addressing clashes
regardless of device placement and without the necessity of a dedicated and
public `BALENA_VPN_GATEWAY` ipv4 address for the remote peer.

Refs: https://tools.ietf.org/html/rfc6598
Connects-to: #67
Change-type: minor

---
##### Contributor checklist
- [X] Tests are included
- [X] Documentation is added or changed
- [X] May effect deployment processes of the component

##### Reviewer Guidelines
---


README.md
```diff
@@ -36,7 +36,8 @@ is used as a DHCP pool for devices.
 Note that the dhcp pool size will also dictate the max clients per
 process, with the max clients per server being
 `max_clients_per_instance * VPN_INSTANCE_COUNT` and not the size of
-the base subnet. A VLSM of `20` will allow for 4094 clients per instance.
+the base subnet. A VLSM of `20` will allow for 4,094 clients per instance, and a
```
dfunckt: This is just a guess, but should `4094` actually be `4096` (2^12) instead?

PR #86, opened 2019-08-01
service: Ensure captureException does not throw

hedss:
The code in the captureException() method 'try's
to get an ID. If it fails, it 'finally' registers
the exception passed to it. Unfortunately, because
it doesn't 'catch', if the 'try' fails, then the
error from it bubbles up and the code calling
'captureException' does not finalise. This is
most prevalent should the API not currently be
running, as it won't register and the timed retry
to register will never get called.

This will also fix other calls made with it.

Change-type: patch
Signed-off-by: Heds Simons <heds@balena.io>

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---

### DIFF
--- a/src/vpn-api/utils/service.ts
+++ b/src/vpn-api/utils/service.ts
@@ -29,9 +29,10 @@ class ServiceInstance {
 		const tags: { [key: string]: string } = {};
 		try {
 			tags.instance_id = this.getId();
-		} finally {
-			captureException(err, fingerprint, { tags });
-		}
+			// tslint:disable-next-line:no-empty
+		} catch {}
+
+		captureException(err, fingerprint, { tags });
 	}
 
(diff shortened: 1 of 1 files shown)


src/vpn-api/utils/service.ts
```diff
@@ -29,9 +29,9 @@ class ServiceInstance {
 		const tags: { [key: string]: string } = {};
 		try {
 			tags.instance_id = this.getId();
-		} finally {
-			captureException(err, fingerprint, { tags });
-		}
+		} catch (_error) {}
```
Page-:
```suggestion
		} catch {}
```

hedss: Done!

PR #88, opened 2019-08-12
device-metrics: create timeseries per-device with device_uuid tag

xginn8:
Connects-to: https://github.com/balena-io/balena-monitor/pull/94
HQ: https://github.com/balena-io/balena/issues/1869
Change-type: minor
Signed-off-by: Matthew McGinn <matthew@balena.io>


xginn8:
generally inserting label names directly into the metric names is discouraged, however in this case we are covered by the one exception that makes sense (from https://www.robustperception.io/on-the-naming-of-things):

> There's one very rare case where this is okay, and that's where you need to distinguish multiple metrics based on the same events but which have different labels for performance reasons.

PR #91, opened 2019-08-23
add prometheus node_exporter service in-image

dt-rush:
re: https://www.flowdock.com/app/rulemotion/resin-tech/threads/3c8dMtg98N1c-PBsR6xJgnpeR3m

quick test if you don't want to run it through BoB:

```
$ docker build -t open-balena-vpn:local .

$ docker run -d --rm -v /sys/fs/cgroup:/sys/fs/cgroup:ro,rslave --security-opt seccomp=unconfined -e 'CONFD_BACKEND=ENV' -e 'BALENA_MONITOR_SECRET_TOKEN=foobar' -it open-balena-vpn:local

$ CONTAINER_IP=$(docker inspect $(docker ps | grep "open-balena-vpn:local" | awk '{print $1}') -f '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}')

$ curl -v --user monitor:foobar ${CONTAINER_IP}:9003/metrics
```

Change-type: minor
Signed-off-by: Nick Payne <nickp@balena.io>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,15 +11,43 @@ COPY src /usr/src/app/src
 RUN npm run build
 
 
+FROM golang:alpine AS gobuild
+
+RUN apk add make curl git && \
+	git clone https://github.com/prometheus/node_exporter && \
+	cd node_exporter && \
+	git checkout 5d3e2ce2ef927eeb44fc409d3cd1fbba884571f3 && \
+	make build && \
+	cp node_exporter /usr/bin/
+
+
 FROM base as main
 
-EXPOSE 80 443 3128
+EXPOSE 80 443 3128 9003
 
 RUN apt-get update -qq \
 	&& apt-get install -qy haproxy=1.8.* iptables --no-install-recommends \
 	&& apt-get clean \
 	&& rm -rf /var/lib/apt/lists/* /etc/apt/sources.list.d/*.list /etc/haproxy/* /etc/openvpn/* /etc/rsyslog.d/49-haproxy.conf \
 	&& ln -sf /usr/src/app/openvpn/scripts /etc/openvpn/scripts
+
+# MUSL-LIBC (for running 99%-static node_exporter binary built in gobuild stage)
+RUN wget http://www.musl-libc.org/releases/musl-1.1.10.tar.gz \
+	&& tar -xvf musl-1.1.10.tar.gz \
+	&& rm musl-1.1.10.tar.gz \
+	&& cd musl-1.1.10 \
+	&& ./configure \
+	&& make \
+	&& make install \
+	&& cd .. \
+	&& rm -rf musl-1.1.10
+
+# install nginx for auth-protecting node exporter metrics path
+RUN apt-get update \
+	&& apt-get -y -f install nginx \
+	&& rm -rf /var/lib/apt/lists/* \
+	&& mkdir -p /etc/nginx
+COPY config/nginx/ /etc/nginx/
 
 ENV LIBNSS_OPENVPN_VERSION 22feb11322182f6fd79f85cd014b65b6c40b7b47
 RUN tmp="$(mktemp -d)" set -x \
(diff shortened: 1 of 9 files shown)


config/nginx/nginx.conf
```diff
@@ -0,0 +1,19 @@
+http {
+
+    server {
+
+        listen 9003;
```
xginn8: why port `9003` here? can we use instead the standard `9100` as defined by prometheus (https://github.com/prometheus/prometheus/wiki/Default-port-allocations), or is this to avoid `:9100` conflicts with other services?

Dockerfile
```diff
@@ -35,9 +63,15 @@ COPY package.json package-lock.json /usr/src/app/
 RUN npm ci --unsafe-perm --production && npm cache clean --force 2>/dev/null
 
 COPY --from=builder /usr/src/app/build /usr/src/app/build
+COPY --from=gobuild /usr/bin/node_exporter /bin/
```
xginn8: i think you should copy this into `/usr/bin`, to stay closer to https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard

config/services/node-exporter.service
```diff
@@ -0,0 +1,51 @@
...
+    --no-collector.textfile \
+    --no-collector.time \
+    --no-collector.timex \
+    --no-collector.uname \
+    --no-collector.vmstat \
+    --no-collector.xfs \
+    --no-collector.zfs
+
+Restart=always
+SyslogIdentifier=node-exporter
```
xginn8: i dunno how this propagates out, but it may be worth noting that this instance specifically is running within the vpn container

config/services/node-exporter.service
```diff
@@ -0,0 +1,51 @@
...
+
+[Service]
+StandardOutput=journal+console
+StandardError=journal+console
+Type=simple
+
+ExecStart=/bin/node_exporter \
+    --web.listen-address :9000 \
+    --web.telemetry-path /metrics \
+    --no-collector.arp \
```
xginn8: since you're disabling all collectors except one, can you explicitly enable that as well to make this easier to understand?

Dockerfile
```diff
@@ -11,16 +11,44 @@ COPY src /usr/src/app/src
...
+	&& tar -xvf musl-1.1.10.tar.gz \
+	&& rm musl-1.1.10.tar.gz \
+	&& cd musl-1.1.10 \
+	&& ./configure \
+	&& make \
+	&& make install \
+	&& cd .. \
+	&& rm -rf musl-1.1.10
+
+# install nginx for auth-protecting node exporter metrics path
```
wrboyce: haproxy is already configured to handle this, you can just add the node_exporter backend [here](https://github.com/balena-io/open-balena-vpn/blob/master/config/confd/templates/haproxy.cfg.tmpl#L36-L60).

Dockerfile
```diff
@@ -11,16 +11,44 @@ COPY src /usr/src/app/src
 RUN npm run build
 
 
+FROM golang:alpine AS gobuild
```
wrboyce: There is no need to build this, we can just use a similar to how we install sshproxy:

```Dockerfile
ENV NODE_EXPORTER_VERSION 0.18.1
ENV NODE_EXPORTER_SHA256SUM b2503fd932f85f4e5baf161268854bf5d22001869b84f00fd2d1f57b51b72424
RUN NODE_EXPORTER_TGZ="/tmp/node_exporter.tar.gz" set -x \
    && curl -Lo "${NODE_EXPORTER_TGZ}" https://github.com/prometheus/node_exporter/releases/download/v${NODE_EXPORTER_VERSION}/node_exporter-${NODE_EXPORTER_VERSION}.linux-amd64.tar.gz \
    && echo "${NODE_EXPORTER_SHA256SUM}  ${NODE_EXPORTER_TGZ}" | sha256sum -c \
    && tar -xzC /usr/local/bin -f "${NODE_EXPORTER_TGZ}" --strip-components=1 --wildcards '*/node_exporter' \
    && rm "${NODE_EXPORTER_TGZ}"
```

Dockerfile
```diff
@@ -11,16 +11,44 @@ COPY src /usr/src/app/src
...
-EXPOSE 80 443 3128
+EXPOSE 80 443 3128 9003
 
 RUN apt-get update -qq \
 	&& apt-get install -qy haproxy=1.8.* iptables --no-install-recommends \
 	&& apt-get clean \
 	&& rm -rf /var/lib/apt/lists/* /etc/apt/sources.list.d/*.list /etc/haproxy/* /etc/openvpn/* /etc/rsyslog.d/49-haproxy.conf \
 	&& ln -sf /usr/src/app/openvpn/scripts /etc/openvpn/scripts
 
+# MUSL-LIBC (for running 99%-static node_exporter binary built in gobuild stage)
```
wrboyce: I don't think this is necessary.

dt-rush:
superceded by #93

PR #94, opened 2019-08-31
proxy: refuse to forward tunnels via self and detect forwarding loops

wrboyce:
Change-type: minor

---
##### Contributor checklist

- [X] Tests are included

##### Reviewer Guidelines
---

### DIFF
--- a/src/proxy-worker.ts
+++ b/src/proxy-worker.ts
@@ -96,6 +96,17 @@ class Tunnel extends nodeTunnel.Tunnel {
 										'device is not available on registered service instance',
 									);
 								}
+								const forwardSignature = `By=open-balena-vpn(${this.serviceId})`;
+								if (req.headers.forwarded != null) {
+									if (req.headers.forwarded.includes(forwardSignature)) {
+										throw new errors.HandledTunnelingError(
+											'loop detected forwarding tunnel request',
+										);
+									}
+									req.headers.forwarded = `${req.headers.forwarded},${forwardSignature}`;
+								} else {
+									req.headers.forwarded = forwardSignature;
+								}
 								this.logger.info(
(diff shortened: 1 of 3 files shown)


src/proxy-worker.ts
```diff
@@ -96,6 +96,18 @@ class Tunnel extends nodeTunnel.Tunnel {
 										'device is not available on registered service instance',
 									);
 								}
+								const forwardSignature = `By=open-balena-vpn(${this.serviceId})`;
+								let forwardHeader: string[] = [];
+								if (req.headers.forwarded != null) {
+									forwardHeader = req.headers.forwarded.split(',');
+									if (forwardHeader.indexOf(forwardSignature) !== -1) {
```
Page-:
```suggestion
									if (forwardHeader.includes(forwardSignature)) {
```

src/proxy-worker.ts
```diff
@@ -96,6 +96,18 @@ class Tunnel extends nodeTunnel.Tunnel {
 										'device is not available on registered service instance',
 									);
 								}
+								const forwardSignature = `By=open-balena-vpn(${this.serviceId})`;
+								let forwardHeader: string[] = [];
+								if (req.headers.forwarded != null) {
```
Page-: It probably makes more sense to just do:
```suggestion
								if (req.headers.forwarded != null) {
									if (req.headers.forwarded.includes(forwardSignature)) {
										throw new errors.HandledTunnelingError(
											'loop detected forwarding tunnel request',
										);
									}
									req.headers.forwarded = `${req.headers.forwarded},${forwardSignature}`
								} else {
									req.headers.forwarded = forwardSignature
								}
```
to avoid the need for creating so many arrays that need to be garbage collected

PR #98, opened 2019-09-27
logging improvements

wrboyce:
Change-type: patch

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---


src/app.ts
```diff
@@ -53,6 +53,8 @@ const logger = getLogger('vpn');
 const VPN_INSTANCE_COUNT =
 	parseInt(process.env.VPN_INSTANCE_COUNT!, 10) || os.cpus().length;
 
+const VPN_VERBOSE_LOGS = process.env.DEFAULT_VERBOSE_LOGS! === 'true';
```
Page-: Is the `!` actually necessary?

wrboyce: good question.

wrboyce: Nope.

src/utils/openvpn.ts
```diff
@@ -268,8 +269,19 @@ export class VpnManager extends EventEmitter {
...
+		this.process.stdout.pipe(es.split()).pipe(
+			es.map((data: string) => {
+				this.emit('log', VpnLogLevels.n, data);
+			}),
+		);
+		this.process.stderr.pipe(es.split()).pipe(
+			es.map((data: string) => {
+				this.emit('log', VpnLogLevels.W, data);
+			}),
+		);
```
petrosagg: Is this stream in flowing mode? We have to make sure it doesn't buffer forever

petrosagg: If you're not going to do further processing it's better to do a `on('data', ..)` handler after you pipe to es.split. This will make the stream flowing and you don't have to use `es.map`.

However, this is not a good way of handling logs because there is no backpressure mechanism. If for whatever reason we get flooded with logs it will just accumulate in RAM.

Not for this PR, but logs should be normal streamed piped end-to-end without going through simple event emitters.

PR #99, opened 2019-09-27
Remove redundant code

Page-:
Change-type: patch

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---

### DIFF
--- a/src/utils/clients.ts
+++ b/src/utils/clients.ts
@@ -111,7 +111,6 @@ const setDeviceState = (() => {
 		} else {
 			deviceStates[uuid].targetState = state;
 		}
-		deviceStates[uuid].targetState = state;
 		return applyState(serviceId, uuid);
 	};
 })();


src/utils/clients.ts
```diff
@@ -111,7 +111,6 @@ const setDeviceState = (() => {
 		} else {
 			deviceStates[uuid].targetState = state;
 		}
-		deviceStates[uuid].targetState = state;
```
dfunckt: I think you need to remove the “else” block instead

dfunckt: Never mind, confused.

Page-: Nope because in the `if` block we create the object with `{..., targetState: state }` so it's only in the existing object case (the else) that we need to update it.  It doesn't technically break anything to do this redundant assignment, but it is confusing and unnecessary

PR #101, opened 2019-09-30
api: use single auth api server with shared pool

wrboyce:
Change-type: minor

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---


src/app.ts
```diff
@@ -95,79 +97,89 @@ if (cluster.isMaster) {
...
+							metrics.reset(Metrics.SessionTxBitrate);
+						})
+						.catch((err: Error) => {
+							serviceLogger.warning(`error in /cluster_metrics: ${err}`);
+							res.status(500).send();
+						});
+				})
+				.listen(8080);
+
+			return Promise.all([app, metrics]);
```
Page-: Aren't both of those synchronous values?  In which case this would be better as:
```suggestion
			return [app, metrics];
```

wrboyce: Good point, I was thinking about the `listenAsync` return value.

src/vpn-worker.ts
```diff
@@ -117,11 +114,11 @@ const worker = (instanceId: number, serviceId: number) => {
 	});
 
 	vpn.on('process:exit', (code, signal) => {
-		let msg = 'process exited';
+		let msg = `openvpn process (pid=${vpn.pid()})`;
 		if (code != null) {
-			msg = `process exited with code ${code}`;
+			msg = `${msg} exited (code=${code})`;
```
Page-: This could also be
```suggestion
			msg += ` exited (code=${code})`;
```

src/vpn-worker.ts
```diff
@@ -117,11 +114,11 @@ const worker = (instanceId: number, serviceId: number) => {
 	});
 
 	vpn.on('process:exit', (code, signal) => {
-		let msg = 'process exited';
+		let msg = `openvpn process (pid=${vpn.pid()})`;
```
Page-: This will now be the message in the case that no code/signal is emitted.  I've got no idea if that's ever actually the case but if so it's probably worth keeping the verb of what happened

wrboyce: Nice catch, fixed.

src/vpn-worker.ts
wrboyce: Is using `+=` better?

Page-: Just that it could make it clearer we're appending to the message.  It also arguably might allow v8 to optimise the appending in ways it won't catch with the other, but that's just a function of how clever the optimiser is at this point in time (and this piece of code doesn't care about performance)

PR #103, opened 2019-10-07
api: filter client state updates by worker id

wrboyce:
Change-type: patch

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -46,6 +46,8 @@ const fromLocalHost: express.RequestHandler = (req, res, next) => {
 export const apiFactory = (serviceId: number) => {
 	const api = express.Router();
 
+	const workerMap: _.Dictionary<string> = {};
+
 	const logger = getLogger('vpn', serviceId);
 
 	const logStateUpdate = (state: clients.DeviceState) => {
@@ -64,6 +66,7 @@ export const apiFactory = (serviceId: number) => {
 		}
 		metrics.inc(Metrics.OnlineDevices);
 		metrics.inc(Metrics.TotalDevices);
+		workerMap[req.body.common_name] = req.params.worker;
 		clients
 			.connected(serviceId, req.params.worker, req.body)
 			.then(logStateUpdate);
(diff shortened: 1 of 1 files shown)


src/api.ts
```diff
@@ -46,6 +46,8 @@ const fromLocalHost: express.RequestHandler = (req, res, next) => {
 export const apiFactory = (serviceId: number) => {
 	const api = express.Router();
 
+	const workerMap: { [key: string]: string } = {};
```
richbayliss:
```suggestion
	const workerMap: _.Dictionary<string> = {};
```

richbayliss: You have `lodash` in your deps anyway, so makes sense to use it's `Dictionary<>` type 👍

src/api.ts
```diff
@@ -64,6 +66,7 @@ export const apiFactory = (serviceId: number) => {
 		}
 		metrics.inc(Metrics.OnlineDevices);
 		metrics.inc(Metrics.TotalDevices);
+		workerMap[req.params.common_name] = req.params.worker;
```
richbayliss: Where does `common_name` come from in the `req.params`?

wrboyce: [this](https://github.com/balena-io/open-balena-vpn/blob/master/openvpn/scripts/client-connect.sh#L26-L33) is called by openvpn when a client connects.

src/api.ts
```diff
@@ -109,6 +112,10 @@ export const apiFactory = (serviceId: number) => {
 			return res.sendStatus(400);
 		}
 
+		if (workerMap[req.body.common_name] !== req.params.worker) {
+			return res.sendStatus(400);
```
pdcastro: Ignore this comment if it is not relevant -- I've jumped in this PR without much context / knowledge of how the VPN service works. But it looks like workerMap is a simple object / dictionary local to the server handling the request. If the requests for `api.post('/api/v2/:worker/clients/')` and `api.delete('/api/v2/:worker/clients/')` were served by different machines (load balancing), then the delete request would be answered with 400.

wrboyce: @pdcastro almost, the `worker id` refers to a process on the machine, not a different physical machine.

pdcastro: But if there were multiple physical machines (maybe there aren't), the issue could still happen... Anyway, it sounds like you're well aware of it, so approving the PR. :-)

src/api.ts
```diff
@@ -109,6 +112,10 @@ export const apiFactory = (serviceId: number) => {
 			return res.sendStatus(400);
 		}
 
+		if (workerMap[req.body.common_name] !== req.params.worker) {
```
Page-: Is it worth adding a sentry exception here so we can see if this happens more than expected? (at least in the short-term)

Page-: Should the workerMap entry be cleared after we've triggered the disconnect?

wrboyce: Great idea, done!

wrboyce: Certainly can't hurt, done.

PR #104, opened 2019-10-07
vpn: kill orphan server processes on startup

wrboyce:
Change-type: patch

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---

### DIFF
--- a/src/utils/openvpn.ts
+++ b/src/utils/openvpn.ts
@@ -80,8 +80,10 @@ const VpnLogLevels = {
 } as const;
 type ValueOf<T> = T[keyof T];
 
-const KILL_TIMEOUT = 5 * 1000;
-const STARTUP_TIMEOUT = 5 * 1000;
+const SECONDS = 1000;
+const STARTUP_TIMEOUT = 5 * SECONDS;
+const TERM_TIMEOUT = 5 * SECONDS;
+const KILL_TIMEOUT = TERM_TIMEOUT * 2;
 
 export declare interface VpnManager {
 	on(event: 'process:error', callback?: (err: Error) => void): this;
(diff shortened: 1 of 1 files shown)


src/utils/openvpn.ts
```diff
@@ -267,7 +272,18 @@ export class VpnManager extends EventEmitter {
 		return emitter;
 	};
 
+	private killOrphan = () => {
+		if (fs.existsSync(this.pidFile)) {
+			const pid = fs
+				.readFileSync(this.pidFile)
+				.toString()
```
Page-:
```suggestion
				.readFileSync(this.pidFile. 'utf8')
```

src/utils/openvpn.ts
```diff
@@ -267,26 +278,53 @@ export class VpnManager extends EventEmitter {
...
+		}
+		const pid = (await fs.readFile(this.pidFile, 'utf8')).replace(/\D/g, '');
+		const signals = { PING: 0, KILL: 9 } as const;
+		const kill = (signal?: ValueOf<typeof signals>) =>
+			spawn('/bin/kill', signal != null ? [`-${signal}`, pid] : [pid]);
+
+		return new Promise(resolve => {
+			const timeout = Date.now() + 5 * 1000;
+			const waitForDeath = (code?: number) => {
+				if (code !== 0) {
```
Page-: If it doesn't equal 0?

src/utils/openvpn.ts
```diff
@@ -19,11 +19,18 @@ import * as Bluebird from 'bluebird';
...
 import * as es from 'event-stream';
 import { EventEmitter } from 'eventemitter3';
+import * as _fs from 'fs';
 import * as net from 'net';
 import VpnConnector = require('telnet-openvpn');
+import { promisify } from 'util';
 
 import { Netmask } from './netmask';
 
+const fs = {
```
Page-: You can use fs.promises btw

src/utils/openvpn.ts
```diff
@@ -267,26 +278,53 @@ export class VpnManager extends EventEmitter {
...
+		if (!(await fs.exists(this.pidFile))) {
+			return;
+		}
+		const pid = (await fs.readFile(this.pidFile, 'utf8')).replace(/\D/g, '');
+		const signals = { PING: 0, KILL: 9 } as const;
+		const kill = (signal?: ValueOf<typeof signals>) =>
+			spawn('/bin/kill', signal != null ? [`-${signal}`, pid] : [pid]);
+
+		return new Promise(resolve => {
+			const timeout = Date.now() + 5 * 1000;
```
Page-: Can you move the 5s to a named constant please

src/utils/openvpn.ts
wrboyce: Yeah, if the signal was _not_ successfully sent (i.e., the process does not exist, it has been killed and our work here is done).

Page-: Ok, can you add that as a comment please.  Also it doesn't cover cases where the issue is in `kill` itself, but that should be pretty rare and we'll just fail later and restart the work/retry the kill

wrboyce: @Page- comment added, and I've also handled the (theoretical) case where a process refuses to term/kill in a timely manner.

src/utils/openvpn.ts
```diff
@@ -267,26 +277,66 @@ export class VpnManager extends EventEmitter {
...
+
+		return new Promise((resolve, reject) => {
+			const start = Date.now();
+			const waitForDeath = (code?: number) => {
+				if (code !== 0) {
+					// the signal was unsuccessful, the pid no longer exists;
+					// the process is dead, and our work here is done.
+					return resolve();
+				}
+				const now = Date.now();
```
Page-: You could make this
```suggestion
				const elapsedTime = Date.now() - start;
```
and then use
```ts
if (elapsedTime  > TERM_TIMEOUT) {
```
which is a little more readable imo

wrboyce: Agreed, will change to this. Cheers.

PR #112, opened 2019-10-29
vpn-worker: convert promise chains to async/await

wrboyce:
Change-type: patch

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---

### DIFF
--- a/src/utils/openvpn.ts
+++ b/src/utils/openvpn.ts
@@ -315,54 +315,51 @@ export class VpnManager extends EventEmitter {
 		});
 	}
 
-	public start() {
-		return Bluebird.resolve(this.killOrphan())
-			.tap(() => {
-				this.process = spawn('/usr/sbin/openvpn', this.args(), {
-					stdio: ['ignore', 'pipe', 'pipe'],
-				});
-				// proxy logs from the child process stdout/stderr
-				this.process.stdout!.pipe(es.split()).on('data', data => {
-					this.emit('log', VpnLogLevels.n, data);
-				});
-				this.process.stderr!.pipe(es.split()).on('data', data => {
-					this.emit('log', VpnLogLevels.n, data);
-				});
-				// proxy error events from the child process
-				this.process.on('error', err => {
-					this.emit('process:error', err);
-				});
-				this.process.on('exit', (code, signal) => {
-					this.emit('process:exit', code, signal);
-				});
-			})
-			.then(() => this.waitForStart());
+	public async start() {
+		await this.killOrphan();
+		this.process = spawn('/usr/sbin/openvpn', this.args(), {
+			stdio: ['ignore', 'pipe', 'pipe'],
+		});
+		// proxy logs from the child process stdout/stderr
+		this.process.stdout!.pipe(es.split()).on('data', data => {
+			this.emit('log', VpnLogLevels.n, data);
+		});
+		this.process.stderr!.pipe(es.split()).on('data', data => {
+			this.emit('log', VpnLogLevels.n, data);
+		});
+		// proxy error events from the child process
+		this.process.on('error', err => {
+			this.emit('process:error', err);
+		});
+		this.process.on('exit', (code, signal) => {
+			this.emit('process:exit', code, signal);
+		});
+		await this.waitForStart();
 	}
 
-	private waitForStart(since: number = Date.now()): Bluebird<true> {
-		return new Bluebird((resolve, reject) => {
-			const socket = new net.Socket();
-			const errorHandler = () => {
-				socket.destroy();
-				reject(new Error('socket not ready'));
-			};
-			const readyHandler = () => {
-				socket.end();
-				resolve(true);
-			};
-			socket.on('ready', readyHandler);
-			socket.on('error', errorHandler);
-			socket.on('timeout', errorHandler);
-			socket.connect(this.mgtPort);
-		})
-			.timeout(STARTUP_TIMEOUT - (Date.now() - since))
-			.catch(err => {
-				if (err instanceof Bluebird.TimeoutError) {
-					throw err;
-				}
-				return this.waitForStart(since);
-			})
-			.return(true);
+	private async waitForStart(since: number = Date.now()) {
+		try {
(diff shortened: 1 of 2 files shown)


src/vpn-worker.ts
```diff
@@ -144,39 +144,27 @@ const worker = (instanceId: number, serviceId: number) => {
...
+		logger.notice('starting...');
+		await vpn.start();
+		logger.info(`openvpn process (pid=${vpn.pid()}) started`);
+		// connect to vpn management console, setup bytecount reporting, then release management hold
+		await vpn.connect();
+		await vpn.enableBytecountReporting(VPN_BYTECOUNT_INTERVAL);
+		await vpn.releaseHold();
+		logger.info(`management hold released`);
+		// register as haproxy backend
+		new HAProxy('/var/run/haproxy.sock').register(
```
Page-: This was previously awaited, was that an accident before or should it still be awaited now?

src/utils/openvpn.ts
```diff
@@ -387,22 +385,22 @@ export class VpnManager extends EventEmitter {
...
 		return this.exec('log on all');
 	}
 
-	public enableBytecountReporting(interval: number) {
+	public async enableBytecountReporting(interval: number) {
 		return this.exec(`bytecount ${interval}`);
 	}
 
-	public releaseHold() {
+	public async releaseHold() {
```
Page-: There's no need to mark these as `async` if all they do is call another promise-returning function - as I understand it functionally it will just cause the promise to be wrapped, ie it's equivalent to doing `return Promise.resolve(this.exec('hold release'))`

src/utils/openvpn.ts
```diff
@@ -387,22 +385,22 @@ export class VpnManager extends EventEmitter {
...
 					this.emit('manager:connect');
 				}),
-		).return(true);
+		).return();
 	}
 
-	public exec(command: string): Bluebird<true> {
-		return Bluebird.try(() => this.connector.exec(command)).return(true);
+	public async exec(command: string): Promise<void> {
+		return Bluebird.try(() => this.connector.exec(command)).return();
```
Page-:
```suggestion
		await this.connector.exec(command);
```

src/utils/openvpn.ts
```diff
@@ -377,7 +375,7 @@ export class VpnManager extends EventEmitter {
 		}
 	}
 
-	public connect(): Bluebird<true> {
+	public async connect(): Promise<void> {
 		return Bluebird.try(() =>
```
Page-:
```suggestion
		await Bluebird.try(() =>
```
and ditch the `.return()`

src/utils/openvpn.ts
```diff
@@ -362,7 +360,7 @@ export class VpnManager extends EventEmitter {
 				}
 				return this.waitForStart(since);
 			})
-			.return(true);
+			.return();
```
Page-: `await` and ditch this
```suggestion
```

wrboyce: This seems to be required otherwise we have a `Promise<unknown>` return value.

src/utils/openvpn.ts
```diff
@@ -377,8 +375,8 @@ export class VpnManager extends EventEmitter {
 		}
 	}
 
-	public connect(): Bluebird<true> {
-		return Bluebird.try(() =>
+	public async connect() {
+		await Bluebird.try(() =>
```
Page-: There's no need for this `Bluebird.try` is there? You can do
```suggestion
		await this.connector
			.connect({
				port: this.mgtPort,
				shellPrompt: '',
			})
		this.emit('manager:connect');
```

src/utils/openvpn.ts
Page-: Can you try again? I've got a feeling it was due to the `return` instead of the `await` in the previous version

src/utils/openvpn.ts
```diff
@@ -362,7 +360,7 @@ export class VpnManager extends EventEmitter {
 				}
```
Page-: Can you switch this `.catch` to an awaited try/catch please

src/utils/openvpn.ts
```diff
@@ -377,21 +374,19 @@ export class VpnManager extends EventEmitter {
...
-					this.emit('manager:connect');
-				}),
-		).return(true);
+	public async connect() {
+		await this.connector
+			.connect({
+				port: this.mgtPort,
+				shellPrompt: '',
+			})
+			.then(() => {
```
Page-:
```suggestion
		this.emit('manager:connect');
```

src/utils/openvpn.ts
```diff
@@ -315,54 +315,51 @@ export class VpnManager extends EventEmitter {
...
+				socket.on('ready', readyHandler);
+				socket.on('error', errorHandler);
+				socket.on('timeout', errorHandler);
+				socket.connect(this.mgtPort);
+			}).timeout(STARTUP_TIMEOUT - (Date.now() - since));
+		} catch (err) {
+			if (err instanceof Bluebird.TimeoutError) {
+				throw err;
+			}
+			await this.waitForStart(since);
```
Page-: You can actually get rid of this btw and do
```ts
 const since = Date.now();
while(true) {
	try {
		...
```
instead

src/utils/openvpn.ts
```diff
@@ -315,54 +315,51 @@ export class VpnManager extends EventEmitter {
...
+			this.emit('log', VpnLogLevels.n, data);
+		});
+		// proxy error events from the child process
+		this.process.on('error', err => {
+			this.emit('process:error', err);
+		});
+		this.process.on('exit', (code, signal) => {
+			this.emit('process:exit', code, signal);
+		});
+		return await this.waitForStart();
```
Page-: Since `this.waitForStart()` returns void this would be slightly clearer as
```suggestion
		await this.waitForStart();
```
to avoid the implication that the `waitForStart` return value is meaningful

src/utils/openvpn.ts
wrboyce: Whoops!

PR #116, opened 2020-01-20
utils: cache `canAccess` lookups for 5s

wrboyce:
Change-type: minor

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---


src/utils/device.ts
```diff
@@ -65,11 +66,7 @@ const canAccessDeviceQuery = balenaApi.prepare<{ id: number }>({
 	id: { '@': 'id' },
 	url: `device(@id)/canAccess`,
 });
-export const canAccessDevice = (
-	device: DeviceInfo,
-	port: number,
-	auth?: Buffer,
-) =>
+const $canAccessDevice = (device: DeviceInfo, port: number, auth?: Buffer) =>
```
Page-: I think you should add a custom cache resolve to use just `device.id` from `device` rather than the `is_connected_to_vpn` status as well

Page-: If `auth` is a `Buffer` then I'm not sure how well that'll play with memoization

Page-: @wrboyce ^ ?

wrboyce: I use `normalize` to generate the cache header now, which takes this into account (`args[2] ?? 'guest'`).

PR #123, opened 2020-03-03
parameterize VPN_CONNECT_PROXY_PORT

ab77:
Relates to:
https://github.com/balena-io/helm-charts/pull/59
https://github.com/balena-io/balena-api/pull/2467

Change-type: minor


config/confd/templates/env.tmpl
```diff
@@ -18,7 +18,7 @@ VPN_BASE_MANAGEMENT_PORT=20000
 VPN_API_PORT=30000
 
 VPN_CONNECT_INSTANCE_COUNT={{ if eq (getv "/resin/vpn/production") "true" }}0{{ else }}1{{ end }}
-VPN_CONNECT_PROXY_PORT=3128
```
wrboyce: This is going the change the listening port, is that your intention?

ab77: I only realised after I put this in, when I started working on the `open-balena-api` PR, that I don't need to change this. Though having already done the work, it might make sense to merge it anyway, so that the flexibility is there in the future?

ab77: OK, I've pushed an update where I split the listen and forward ports, since these could be different (in the case of k8s).

config/confd/templates/env.tmpl
```diff
@@ -18,7 +18,8 @@ VPN_BASE_MANAGEMENT_PORT=20000
...
+VPN_CONNECT_PROXY_PORT={{if exists "/resin/vpn/vpn-connect-proxy-port"}}"{{getv "/resin/vpn/vpn-connect-proxy-port"}}"{{else}}{{end}}
```
wrboyce: As far as I can see 3128 is hard-coded in the [helm charts](https://github.com/balena-io/helm-charts/blob/master/balena-vpn/templates/service-nodeport.yaml#L16) so I don't think this needs to be parameterised.

src/proxy-worker.ts
```diff
@@ -23,12 +23,13 @@ import * as nodeTunnel from 'node-tunnel';
...
 import { captureException, device, errors, getLogger, Metrics } from './utils';
+import { intVar } from './utils/config';
 
 const VPN_SERVICE_API_KEY = Buffer.from(process.env.VPN_SERVICE_API_KEY!);
-const VPN_CONNECT_PROXY_PORT = parseInt(
-	process.env.VPN_CONNECT_PROXY_PORT!,
-	10,
-);
+// listen port
+const VPN_CONNECT_PROXY_PORT = intVar('VPN_CONNECT_PROXY_PORT', 3128);
```
wrboyce: I think you can just follow the current convention of using `parseInt` here (or change all other the port params to use `intVar` too, if you'd rather… in which case I'd argue a `portVar` function would be worthwhile with additional validity checks).

ab77: OK, I'll see what I can do.

ab77:
@wrboyce I've made it simpler for now and reverted some of the changes. Hopefully this is a good base to improve from at a later date.

PR #130, opened 2020-04-03
resolve service instance count from podinfo

ab77:
Change-type: minor

### DIFF
--- a/src/utils/config.ts
+++ b/src/utils/config.ts
@@ -1,0 +1,79 @@
+/*
+	Copyright (C) 2020 Balena Ltd.
+
+	This program is free software: you can redistribute it and/or modify
+	it under the terms of the GNU Affero General Public License as published
+	by the Free Software Foundation, either version 3 of the License, or
+	(at your option) any later version.
+
+	This program is distributed in the hope that it will be useful,
+	but WITHOUT ANY WARRANTY; without even the implied warranty of
+	MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+	GNU Affero General Public License for more details.
+
+	You should have received a copy of the GNU Affero General Public License
+	along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import * as fs from 'fs';
+import * as os from 'os';
+
+export const getEnv = <T extends string>(
+	...args: T[]
+): { [key in T]: string } => {
+	args
+		.filter(key => process.env[key] == null)
+		.forEach((key, idx, keys) => {
+			console.log(`${key} env variable is not set.`);
+			if (idx === keys.length - 1) {
+				process.exit(1);
+			}
+		});
+
+	type Env = { [key in T]: string };
+	const env: Partial<Env> = {};
+	args.forEach((key: T) => {
+		env[key] = process.env[key]!;
+	});
+	return env as Env;
+};
+
+// Code copied from our open source API
+// https://github.com/balena-io/open-balena-api/blob/e9abe8f959c59bbeefcadbfdc59642af565b1427/src/lib/config.ts
+export function intVar(varName: string): number;
+export function intVar<R>(varName: string, defaultValue: R): number | R;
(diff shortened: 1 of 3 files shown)


tsconfig.json
```diff
@@ -5,14 +5,14 @@
...
-		"noUnusedLocals": true,
+		"noImplicitAny": true,
 		"noUnusedParameters": true,
+		"noUnusedLocals": true,
 		"outDir": "build/",
 		"preserveConstEnums": true,
 		"removeComments": true,
 		"resolveJsonModule": true,
 		"sourceMap": true,
-		"strict": true,
```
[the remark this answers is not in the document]
ab77: @Page- with `strict: true`, it fails on:

```
src/utils/config.ts(35,15): error TS2345: Argument of type '(key: T) => void' is not assignable to parameter of type '(value: string, index: number, array: string[]) => void'.
  Types of parameters 'key' and 'value' are incompatible.
    Type 'string' is not assignable to type 'T'.
      'string' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint 'string'.
```

src/utils/config.ts
```diff
@@ -0,0 +1,79 @@
...
+
+	You should have received a copy of the GNU Affero General Public License
+	along with this program.  If not, see <https://www.gnu.org/licenses/>.
+*/
+
+import * as fs from 'fs';
+import * as os from 'os';
+
+export const getEnv = <T extends string>(
+	...args: string[]
```
Page-, line 22: Try
```suggestion
	...args: T[]
```
for fixing the issue with `strict`

ab77:
@balena-ci retest

ab77:
@balena-ci retest

ab77:
@balena-ci retest

ab77:
@balena-ci retest

tsconfig.json
```diff
@@ -5,8 +5,9 @@
 		"module": "commonjs",
 		"noErrorTruncation": true,
 		"noFallthroughCasesInSwitch": true,
-		"noUnusedLocals": true,
+		"noImplicitAny": true,
```
wrboyce, line 8: I think this might be implied by `strict=true`?

ab77: I can't find a reference to it under `--strict` [here](https://www.typescriptlang.org/docs/handbook/compiler-options.html) @wrboyce

ab77:
@balena-ci retest

tsconfig.json
```diff
@@ -5,8 +5,8 @@
 		"module": "commonjs",
 		"noErrorTruncation": true,
 		"noFallthroughCasesInSwitch": true,
-		"noUnusedLocals": true,
 		"noUnusedParameters": true,
+		"noUnusedLocals": true,
```
wrboyce, line 9: NIT: these were in alphabetic order.

ab77: This is what `jq` spits out @wrboyce:

```
$ cat tsconfig.json | jq -r --tab .
{
        "compilerOptions": {
                "allowJs": false,
                "forceConsistentCasingInFileNames": true,
                "module": "commonjs",
                "noErrorTruncation": true,
                "noFallthroughCasesInSwitch": true,
                "noUnusedParameters": true,
                "noUnusedLocals": true,
                "outDir": "build/",
                "preserveConstEnums": true,
                "removeComments": true,
                "resolveJsonModule": true,
                "sourceMap": true,
                "strict": true,
                "skipLibCheck": true,
                "strictNullChecks": true,
                "target": "es2018"
        },
        "include": [
                "src/**/*.ts",
                "typings/**/*.d.ts"
        ],
        "exclude": [
                "node_modules"
        ]
}
```

wrboyce: ah, fair enough :+1:

ab77: 🙄

ab77: there is a `-S` parameter to `jq` I just learned about today @wrboyce :)

ab77:
@balena-ci retest

PR #131, opened 2020-04-07
handle zero values correctly

ab77:
Change-type: patch

### DIFF
--- a/src/utils/config.ts
+++ b/src/utils/config.ts
@@ -61,7 +61,7 @@ export function intVar<R>(varName: string, defaultValue?: R): number | R {
 // resolve number of workers based on number of CPUs assigned to pods or available CPUs
 export const getInstanceCount = (varName: string) => {
 	let instanceCount = intVar(varName, null);
-	if (instanceCount != null) {
+	if (instanceCount != null && instanceCount > 0) {
 		return instanceCount;
 	}
 	try {


src/utils/config.ts
```diff
@@ -61,7 +61,7 @@ export function intVar<R>(varName: string, defaultValue?: R): number | R {
 // resolve number of workers based on number of CPUs assigned to pods or available CPUs
 export const getInstanceCount = (varName: string) => {
 	let instanceCount = intVar(varName, null);
-	if (instanceCount != null) {
+	if (instanceCount != null && instanceCount > 0) {
```
Page-, line 64: Why is it set to 0?

wrboyce: 0 means autodetect

Page-: Don't we already have unset for that @wrboyce ?

wrboyce: Not really, `VPN_INSTANCE_COUNT` is required to be set or the app won't start.

Page-: What requires it? And if it's required why are we treating it nullable here? Also what will happen for the openvpn processes if it's set to 0?

wrboyce: > What requires it?

app.ts

> Also what will happen for the openvpn processes if it's set to 0?

one openvpn process is spawned by each worker regardless of this value.

Page-: But app.ts also has handling for it not being set? Why do we have code to handle it not being set if we also require it to be set, it doesn't make sense imo

As for openvpn I realized what I actually meant was the haproxy config

ab77:
@balena-ci retest

ab77:
@balena-ci retest

src/utils/config.ts
ab77: @Page- @wrboyce are we good to merge on this one?

wrboyce: I'm not sure what you're asking about haproxy @Page-? Each worker will register itself as a backend worker with haproxy on startup.

wrboyce: > But app.ts also has handling for it not being set?

I'd argue it has handling for it being set to an invalid value.

Page-: What is the
```
{{- /* servers = getenv(VPN_INSTANCE_COUNT, default=getenv(NPROC) if production else 1) */ -}}
{{ $servers := (getenv "VPN_INSTANCE_COUNT" (or (and (eq (getenv "PRODUCTION_MODE") "true") (getenv "NPROC")) "1")) -}}
```
stuff for in the haproxy config @wrboyce ?

And yeah I think it's fine @ab77, at worst it'll be another PR for haproxy :P

wrboyce: Hmm good point Page, that should actually fall back to `getenv(NPROC)` if `VPN_INSTANCE_COUNT` is <1.

@ab77 is this change actually required here? We set `VPN_INSTANCE_COUNT` to `NPROC` in `env.tmpl` if it isn't explicitly set or if it is set to zero.

```
VPN_INSTANCE_COUNT={{ if gt (getv "/resin/vpn/instances" "0") "0" }}{{ getv "/resin/vpn/instances" }}{{ else }}{{ if eq (getv "/resin/vpn/production") "true" }}{{getenv "NPROC"}}{{ else }}1{{ end }}{{ end }}
```

wrboyce: For reference `NPROC` is set [here](https://github.com/balena-io/open-balena-vpn/blob/master/config/services/confd.service.d/confd.conf).

ab77:
@wrboyce you are correct - VPN instance count auto-detect logic isn't working, because it defaults to NPROC and since in k8s staging, the env var is not set, it eats up 8 CPUs per pod. I am closing this in favour of https://github.com/balena-io/balenacloud-staging/pull/109

PR #133, opened 2020-04-15
refactor instance number autodetect

ab77:
* move logic to getInstanceCount function
* refactor conf.d logic to handle additional cases

Change-type: patch


ab77:
@balena-ci retest

PR #152, opened 2020-10-16
Avoid unintentionally keeping large strings in memory

Page-:
`data.common_name` is constructed from openvpn logs via substrings/slices in such a way that a reference to the original large string is kept around, that string can be anywhere from 1-20kb from the looks of it and when multiplied per each uuid that ever connects it can end up leaking significant memory. `flatstr` purports to handle flattening these strings in a performant method and should be updated as new methods are necessary/etc so makes sense as the first thing to try, but there are alternatives, eg `Buffer.from(s).toString()` should apparently work as well

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1640,6 +1640,11 @@
         "is-buffer": "~2.0.3"
       }
     },
+    "flatstr": {
+      "version": "1.0.12",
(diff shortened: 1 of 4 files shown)


xginn8 (approved):
seems like a good starting point

ab77 (approved):

Page-:
@balena-ci rebase

PR #182, opened 2021-08-25
Remove BALENA_|RESIN_ prefixes

ab77:


ab77:
@dfunckt @Page @pranavpeshwe - not sure who is managing this component, but this is all tested in bob and works as far as I can tell (I've done a E2E browser test and DWB/provision device/etc.)

PR #199, opened 2022-02-15
Remove systemd

robbe-haesendonck:
Removed systemd from open-balena-vpn container to allow running the container unprivileged.
Change-type: minor

---
##### Contributor checklist
- [ ] Tests are included
- [x] Affects the development, build or deployment processes of the component

##### Reviewer Guidelines
---


robbe-haesendonck:
Note: capability NET_ADMIN is still required.

ab77:
@robbe-haesendonck we still require Systemd variant of this component, as it is used in local-dev environments on top of mDNS. If you could update the PR to build both in parallel (using .resinci.yml), it should be OK to merge.

robbe-haesendonck:
@ab77 Is there documentation available for the resinci.yml file?

ab77:
> @ab77 Is there documentation available for the resinci.yml file?

unfortunately not, but for your purposes, something like [this](https://github.com/balena-io-modules/open-balena-base/blob/master/.resinci.yml) should work:

```
---
reviewers: 1
docker:
  builds:
    - path: .
      dockerfile: Dockerfile
      docker_repo: balena/open-balena-vpn
    - path: .
      dockerfile: Dockerfile.no-systemd
      docker_repo: balena/open-balena-vpn:no-systemd
```

robbe-haesendonck:
@ab77 I changed the resinci file as requested and moved the no-systemd version to Dockerfile.no-systemd

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM balena/open-balena-base:v13.4.0 as base
```
ab77, line 1: @robbe-haesendonck I don't believe this and most of the below changes are intentional?

ab77 (changes requested):

Dockerfile
robbe-haesendonck: Yeah, I don't exactly know what went wrong. Be it in my mind.
I'll check/rollback what's necessary when I've got the time.

PR #210, opened 2022-02-28
Only compare the connected state when deciding if we need to report it

Page-:
Since the connected/disconnected status is the only thing that is
important to the api then comparing just that allows us to simplify the
code and ensure we only update when truly necessary

Change-type: patch


Page-:
@balena-ci I self-certify!

Issue #224, opened 2022-04-18
Should limit batching connectivity events to avoid too large POST bodies leading to 413 errors


thgreasi:
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/OhMy8GplWPRfFES56zRO6-0WNh9
See: https://github.com/balena-io/open-balena-vpn/blob/cfa2f6f4f43eac0209407c9e513ac7b997c51387/src/utils/clients.ts#L58-L71

thgreasi:
Resolved by: #225

PR #225, opened 2022-04-21
Limit batching connectivity updates to avoid Payload Too Large errors

thgreasi:
| uuids | content-length | result |
| ----- | -------------- | ------ |
| 8000 | 520043 | Works |
| 8060 | 523943 | Works |
| 8100 | 526543 | Fails |

Connects-to: #224
Change-type: patch
See: https://www.flowdock.com/app/rulemotion/resin-tech/threads/OhMy8GplWPRfFES56zRO6-0WNh9

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---


src/utils/clients.ts
```diff
@@ -50,45 +57,50 @@ export const setConnected = (() => {
...
 			return;
-		} catch (err) {
-			captureException(err, 'device-state-update-error');
-			for (const uuid of uuids) {
-				// If we failed then add the uuids back into the list of pending updates
-				pendingUpdates.add(uuid);
-			}
 		}
+		const eventType = connected ? 'connect' : 'disconnect';
+		const uuidChucks = _.chunk(uuids, API_DEVICE_STATE_POST_BATCH_SIZE);
```
Page-, line 64:
```suggestion
		const uuidChunks = _.chunk(uuids, API_DEVICE_STATE_POST_BATCH_SIZE);
```

src/utils/clients.ts
```diff
@@ -25,15 +25,22 @@
...
+import { request, Response } from './request';
+import { intVar } from './config';
 
 const BALENA_API_HOST = process.env.BALENA_API_HOST!;
 const REQUEST_TIMEOUT = 60000;
+// As of writing this, using a chunk of 8000 62-char UUIDs results a content-length
+// that is bellow the 512KiB threshold that would trigger a 413 http error.
+const API_DEVICE_STATE_POST_BATCH_SIZE = intVar(
+	'API_DEVICE_STATE_POST_BATCH_SIZE',
+	8000,
```
Page-, line 42: Did you test this limit btw @thgreasi ?

src/utils/clients.ts
```diff
@@ -50,45 +57,50 @@ export const setConnected = (() => {
...
-				// If we failed then add the uuids back into the list of pending updates
-				pendingUpdates.add(uuid);
-			}
 		}
+		const eventType = connected ? 'connect' : 'disconnect';
+		const uuidChucks = _.chunk(uuids, API_DEVICE_STATE_POST_BATCH_SIZE);
+		await Promise.allSettled(
+			uuidChucks.map(async (uuidChuck) => {
+				try {
+					const response: Response = await request
```
Page-, line 68: What does typescript infer the type to be?

thgreasi: It was any previously, so in guessing that the request-promise typings are not the best.

src/utils/clients.ts
thgreasi: Yes, I tested it on my BoB @Page-, increasing this till I got a 413. Mentioned the interesting part of my findings in the PR description. 8060 also worked, but I thought I should better be not use the edge value (otherwise we could have a formula to calculate this number).

Page-: I'd actually missed that, and yeah I definitely prefer staying well within a safe margin so lgtm

Page- (approved):

Page-:
@balena-ci republish

PR #229, opened 2022-04-28
Allow the client-connect/disconnect script to fail if necessary

Page-:
These curl commands should never fail as they're talking to the local
nodejs instance, but if for some reason they do it's better to reject
the client than pretend things are ok because they're really not

Change-type: patch

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---

### DIFF
--- a/openvpn/scripts/client-connect.sh
+++ b/openvpn/scripts/client-connect.sh
@@ -21,7 +21,7 @@ set -o errexit
 VPN_INSTANCE_ID=$1
 VPN_API_PORT=$2
 
-curl -s -X POST -H 'Content-type: application/json' -d @- "http://127.0.0.1:${VPN_API_PORT}/api/v2/${VPN_INSTANCE_ID}/clients" >/dev/null <<-EOF || true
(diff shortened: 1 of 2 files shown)


Page-:
@balena-ci I self-certify!

openvpn/scripts/client-connect.sh
```diff
@@ -21,7 +21,7 @@ set -o errexit
 VPN_INSTANCE_ID=$1
 VPN_API_PORT=$2
 
-curl -s -X POST -H 'Content-type: application/json' -d @- "http://127.0.0.1:${VPN_API_PORT}/api/v2/${VPN_INSTANCE_ID}/clients" >/dev/null <<-EOF || true
+curl -s -X POST -H 'Content-type: application/json' -d @- "http://127.0.0.1:${VPN_API_PORT}/api/v2/${VPN_INSTANCE_ID}/clients" >/dev/null <<-EOF
```
pranavpeshwe, line 24: @Page- Can we remove the `/dev/null` as well?

PR #232, opened 2022-05-17
Fixes check for uuid to serviceID assignement

fisehara:
As the vpn registers a serviceID with UUIDs in the api
checking back if the serviceID is registered with the UUID should not throw an error on equality.

Change-type: patch

### DIFF
--- a/src/proxy-worker.ts
+++ b/src/proxy-worker.ts
@@ -91,7 +91,7 @@ class Tunnel extends nodeTunnel.Tunnel {
 				return socket;
 			} else {
 				const vpnHost = await device.getDeviceVpnHost(uuid, auth);
-				if (vpnHost.id === this.serviceId) {
+				if (vpnHost.id !== this.serviceId) {
 					client.end(HTTP_500);
 					throw new errors.HandledTunnelingError(
 						'device is not available on registered service instance',


src/proxy-worker.ts
```diff
@@ -91,7 +91,7 @@ class Tunnel extends nodeTunnel.Tunnel {
 				return socket;
 			} else {
 				const vpnHost = await device.getDeviceVpnHost(uuid, auth);
-				if (vpnHost.id === this.serviceId) {
+				if (vpnHost.id !== this.serviceId) {
```
pranavpeshwe, line 94: I fear this will break everything @fisehara . By everything, I mean to say that all devices will get a `500`. On line 80, we check if the device is served by the VPN instance running this code. In that case `vpnHost.id === this.serviceId` will be true by design. In the `else` clause, `vpnHost.id` should never be equal to `this.serviceId`. If they are equal (which is the original check on line 94), then it means the system is not consistent which is why we return a `500` in that case.
The fact that such a check was introduced implies that this case did arise (or was envisioned by whoever wrote the code). @Page- , do you happen to remember/know why this was done? It seems that our mechanism of relaying device connectivity/disconnectivity can fail to happen (or get delayed). Pods getting recycled is one scenario that comes to mind.

Page-:
The code as it is currently is correct. It is checking which VPN host the device is marked as being connected and if it is the host we are currently on then we error as it is not actually available (since we look for the device locally before ever talking to the api) and if it is not on the current host then we forward the request to the correct host.
The change as proposed would invert that and if the device is expected to be on another vpn host then we'll reject it (and the client will have to keep trying until it ends up on the correct instance by pot luck, which could take a long time) and if it is meant to be on the current host it'll instead try to forward to the current instance and end up in an infinite recursion

PR #239, opened 2022-05-25
Minimize time spent responding to openvpn scripts

Page-:
Some of these can be blocking so the faster we respond the faster
openvpn will continue processing other clients

Change-type: patch

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -36,7 +36,7 @@ import { hasDurationData, isTrusted } from './utils/openvpn';
 const fromLocalHost: express.RequestHandler = (req, res, next) => {
 	// '::ffff:127.0.0.1' is the ipv4 mapped ipv6 address and ::1 is the ipv6 loopback
 	if (!['127.0.0.1', '::ffff:127.0.0.1', '::1'].includes(req.ip)) {
-		return res.sendStatus(401);
+		return res.status(401).end();
 	}
 
 	next();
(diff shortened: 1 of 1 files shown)


src/api.ts
```diff
@@ -125,14 +127,15 @@ export const apiFactory = (serviceId: number) => {
 				'openvpn-oos-event',
 				{ tags: { uuid }, req },
 			);
-			return res.sendStatus(400);
+			return res.status(400).end();
 		}
+		res.status(200).end();
```
Page-, line 132: Moving this before the `metrics.dec`/`clients.setConnected` is the biggest win here

src/api.ts
```diff
@@ -53,8 +53,11 @@ export const apiFactory = (serviceId: number) => {
 
 	api.post('/api/v2/:worker/clients/', fromLocalHost, (req, res) => {
 		if (!isTrusted(req.body)) {
-			return res.sendStatus(400);
+			return res.status(400).end();
 		}
+		// Immediately respond to minimize time in the client-connect script
+		res.status(200).end();
```
Page-, line 59: This is the biggest win here as it means we immediately respond to the client-connect script which definitely blocks, the other changes are just a general improvement by removing unnecessary body messages and such

Page-:
@balena-ci I self-certify!

Page-:
@balena-ci retest

Page-:
@balena-ci republish

Page-:
@balena-ci republish

PR #258, opened 2022-06-23
Don't restart workers while draining

ab77:
Change-type: patch

### DIFF
--- a/src/vpn-worker.ts
+++ b/src/vpn-worker.ts
@@ -105,72 +105,62 @@ const worker = async (instanceId: number, serviceId: number) => {
 	});
 
 	const drainConnections = async () => {
-		const clientCount = Object.keys(clientCache).length;
-		logger.info(`${clientCount} clients connected`);
+		const drainQueue = Object.values(clientCache);
+		const clientCount = drainQueue.length;
 
 		if (clientCount > 0) {
-			logger.debug(JSON.stringify(clientCache));
+			logger.info(
+				`attempt to drain ${clientCount} connected clients in ${DEFAULT_SIGTERM_TIMEOUT}ms`,
+			);
 			const delayMs = DEFAULT_SIGTERM_TIMEOUT / clientCount;
 			logger.info(
 				`disconnecting ${clientCount} clients, spaced by ${delayMs}ms`,
 			);
 
-			let timeToKill = DEFAULT_SIGTERM_TIMEOUT;
-
-			for (let clientId = 0; clientId < clientCount; clientId++) {
+			for (const client of drainQueue) {
 				try {
-					logger.debug(JSON.stringify(clientCache[clientId]));
-					const cn = clientCache[clientId].uuid;
-					logger.info(`disconnecting ${cn}`);
-					// update device state
-					clients.setConnected(cn, serviceId, false, logger);
-					// disconnect client from VPN
-					await vpn.killClient(cn);
+					const cn = client.uuid;
+					try {
+						logger.info(`disconnecting ${cn}`);
+						// update device state
+						clients.setConnected(cn, serviceId, false, logger);
+						// disconnect client from VPN
+						await vpn.killClient(cn);
+					} catch (err) {
+						logger.warning(`${err} while killing ${cn} on worker ${serviceId}`);
+					}
 				} catch (err) {
 					logger.warning(`received '${err}' trying to disconnect client`);
 				}
-				// last client disconnected
-				if (clientCount - clientId === 1) {
-					logger.info(
-						`all ${clientCount} client(s) disconnected, signalling cluster`,
-					);
-					// signal drain status to master
-					if (typeof process.send === 'function') {
-						process.send({
-							type: 'drain',
-							data: {
-								instanceId,
-								finished: true,
-							},
-						});
-					}
-				}
 				// otherwise keep disconnecting clients
 				await setTimeout(delayMs);
-				timeToKill = timeToKill - delayMs;
-			}
-		} else {
-			logger.info(`${clientCount} clients connected, signalling cluster`);
-			// signal drain status to master
-			if (typeof process.send === 'function') {
-				process.send({
-					type: 'drain',
-					data: {
-						instanceId,
-						finished: true,
-					},
-				});
 			}
 		}
+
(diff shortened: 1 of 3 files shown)


ab77:
@balena-ci retest

ab77:
@balena-ci retest

ab77:
@balena-ci retest

ab77:
@balena-ci retest

src/vpn-worker.ts
```diff
@@ -105,32 +105,45 @@ const worker = async (instanceId: number, serviceId: number) => {
...
-			logger.debug(JSON.stringify(clientCache));
 			const delayMs = DEFAULT_SIGTERM_TIMEOUT / clientCount;
 			logger.info(
 				`disconnecting ${clientCount} clients, spaced by ${delayMs}ms`,
 			);
 
 			let timeToKill = DEFAULT_SIGTERM_TIMEOUT;
 
-			for (let clientId = 0; clientId < clientCount; clientId++) {
+			for (const [key] of Object.entries(clientCache)) {
```
Page-, line 119: If you only want the key you're better off doing
```suggestion
			for (const key of Object.keys(clientCache)) {
```
but in this case I think you actually only care about the value, ie
```suggestion
			for (const cn of Object.values(clientCache)) {
```
and can remove all of the dancing around string vs number indexing below

src/vpn-worker.ts
```diff
@@ -105,32 +105,45 @@ const worker = async (instanceId: number, serviceId: number) => {
...
+						// https://stackoverflow.com/questions/3633362/is-there-any-way-to-use-a-numeric-type-as-an-object-key/3633390#3633390
+						// https://bobbyhadz.com/blog/typescript-element-implicitly-any-type-not-number
+						// https://www.typescriptlang.org/tsconfig#suppressImplicitAnyIndexErrors
+					}
 				} catch (err) {
 					logger.warning(`received '${err}' trying to disconnect client`);
 				}
 				// last client disconnected
-				if (clientCount - clientId === 1) {
+				if (clientCount <= 1) {
```
Page-, line 146: You should be able to just move this outside of the `for` loop and ignore client counting since you'll have gone through and attempted disconnections of them all at that point, ie they'll all have had a disconnect triggered

src/vpn-worker.ts
Page-: I assume you were using `Object.values` and then trying to use that to something that expected a string? I'm assuming it's fixed now because I would expect the current version to work?

src/vpn-worker.ts
```diff
@@ -105,63 +105,62 @@ const worker = async (instanceId: number, serviceId: number) => {
 	});
 
 	const drainConnections = async () => {
-		const clientCount = Object.keys(clientCache).length;
+		let clientCount = Object.keys(clientCache).length;
```
Page-, line 108: I think it makes more sense to grab the values here
```suggestion
		const clients = Object.values(clientCache);
```
and then use either `clients.length` or `const clientCount = clients.length` and also reuse it as
```
for (const cn of clients) {
```
below to avoid repeating the work

src/vpn-worker.ts
```diff
@@ -105,63 +105,62 @@ const worker = async (instanceId: number, serviceId: number) => {
...
+								`${err} while killing ${cn} on worker ${serviceId}`,
+							);
+						}
+
+						// update device state
+						clients.setConnected(cn, serviceId, false, logger);
+					} catch {
+						// https://stackoverflow.com/questions/3633362/is-there-any-way-to-use-a-numeric-type-as-an-object-key/3633390#3633390
+						// https://bobbyhadz.com/blog/typescript-element-implicitly-any-type-not-number
+						// https://www.typescriptlang.org/tsconfig#suppressImplicitAnyIndexErrors
```
Page-, line 137: These seem to be related to the previous issue around using the object key from `Object.entries` so I'm assuming these comments are no longer necessary? (and probably the try/catch is also no longer necessary?)

src/vpn-worker.ts
```diff
@@ -105,63 +105,62 @@ const worker = async (instanceId: number, serviceId: number) => {
...
-			// signal drain status to master
-			if (typeof process.send === 'function') {
-				process.send({
-					type: 'drain',
-					data: {
-						instanceId,
-						finished: true,
-					},
-				});
+				clientCount--;
```
Page-, line 145:
```suggestion
```

src/vpn-worker.ts
ab77: So far the only way I've been able to squelch these compiler errors is to use what Harrold suggested:
```
                       for (const [key] of Object.entries(clientCache)) {
                           try {
                               const _key = parseInt(key, 10);
```

ab77: I think I got it..

ab77:
@balena-ci retest

src/app.ts
```diff
@@ -146,6 +146,10 @@ if (cluster.isPrimary) {
 				}`,
 			);
 			_.times(VPN_INSTANCE_COUNT, (i) => {
+				const drainCount = Object.keys(workerStates).length;
```
ab77, line 149: @Page- do you happen to know how `drainCount` is resolved here?

I always get:

```
Jun 27 23:18:01 open-balena-vpn-589468c9ff-zr4kz vpn[530]: info: [master-50933] 0/4 instances draining
Jun 27 23:18:01 open-balena-vpn-589468c9ff-zr4kz vpn[530]: info: [master-50933] 0/4 instances draining
Jun 27 23:18:01 open-balena-vpn-589468c9ff-zr4kz vpn[530]: info: [master-50933] 0/4 instances draining
Jun 27 23:18:01 open-balena-vpn-589468c9ff-zr4kz vpn[530]: info: [master-50933] 0/4 instances draining
```

... which suggests a copy of `workerStates` is taken at init only. If that's the case, do I need to pass a pointer to it somehow to the service.wrap() function?

ab77: I've changed tactic a bit and trying to route every interrupt in the worker to the drain function, including the `unhandledException`.

ab77:
@balena-ci I self-certify!

ab77:
@balena-ci republish

PR #262, opened 2022-06-30
Handle uncaughtException(s)

ab77:
* prevent restart loops during draining
* check drain process is already in progress
* attempt to drain on unhandled exceptions in worker(s) instead of dropping all clients immediately

Change-type: patch


ab77:
@balena-ci retest

ab77:
@balena-ci I self-certify!

PR #269, opened 2022-07-04
handle uncaughtException(s) in VpnManager

ab77:
Change-type: patch


ab77:
@balena-ci retest

ab77:
@balena-ci retest

ab77:
@balena-ci retest

ab77:
@balena-ci retest

src/utils/openvpn.ts
```diff
@@ -83,6 +83,10 @@ const TERM_TIMEOUT = 5 * SECONDS;
 const KILL_TIMEOUT = TERM_TIMEOUT * 2;
 
 export declare interface VpnManager {
+	on(
```
ab77, line 86: @Page- so I am guessing now this is superfluous?

src/vpn-worker.ts
```diff
@@ -182,6 +177,11 @@ const worker = async (instanceId: number, serviceId: number) => {
 		verbose,
 	);
 
+	vpn.on('process:uncaughtException', async (err, origin) => {
```
ab77, line 180: @Page- this is superfluous also?

Page-: Yes, and the correct way to catch it would actually be
```suggestion
	process.on('uncaughtException', async (err, origin) => {
```
see: https://nodejs.org/docs/latest-v16.x/api/process.html#event-uncaughtexception

The `process:error`/`process:exit` errors the `VpnManager` interface has are for the child_process interface to the openvpn process, whereas the `uncaughtException` is on the nodejs process

src/vpn-worker.ts
```diff
@@ -102,67 +102,65 @@ const worker = async (instanceId: number, serviceId: number) => {
 		}
 	});
 
-	let draining = false;
```
Page-, line 105: I think only allowing one draining attempt makes sense, but you could also do it via
```ts
const drainConnections = _.once(async () => {
```
if you want to simplify it (it won't have any logging for it though)

src/vpn-worker.ts
```diff
@@ -102,67 +102,65 @@ const worker = async (instanceId: number, serviceId: number) => {
...
+				type: 'drain',
+				data: {
+					instanceId,
+					finished: true,
+				},
+			});
 		}
+
+		// wait here, the master should exit when all workers have signaled completion
+		await setTimeout(DEFAULT_SIGTERM_TIMEOUT);
```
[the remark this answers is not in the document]
ab77: @Page- yes, this is the reason why. I wasn't able to figure out how to do this in the master process. I spent a bit of time modifying the logic [here](https://github.com/balena-io/open-balena-vpn/blob/master/src/app.ts#L148-L168), but I couldn't get it to work.

ab77:
@balena-ci retest

src/vpn-worker.ts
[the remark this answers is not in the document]
ab77: maybe a global variable?

ab77:
@balena-ci retest

src/vpn-worker.ts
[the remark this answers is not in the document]
ab77: otherwise I've tested it and the draining appears to work.

pranavpeshwe (commented):
> Co-authored-by:

Just wanted to mention that this is the first time I've seen ^^. Like the spirit 👍🏻 👍🏻

src/vpn-worker.ts
```diff
@@ -102,68 +103,67 @@ const worker = async (instanceId: number, serviceId: number) => {
 		}
 	});
 
-	let draining = false;
-	const drainConnections = async () => {
+	const drainConnections = _.once(async () => {
```
pranavpeshwe, line 106: Given that this piece of code plays a critical role, could we have a small comment here describing what the function does and why? More specifically - are there any caveats and/or assumptions (such as any k8s settings) that our future selves should be reminded of?

pranavpeshwe: Going by the FD thread, it is clear that quite some thought & discussion has gone into it.

src/vpn-worker.ts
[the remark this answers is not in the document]
Page-: @ab77 I think you could probably do something like
```
			let terminating = false;
			process.on('SIGTERM', () => {
				terminating = true;
			});
			_.times(VPN_INSTANCE_COUNT, (i) => {
				const workerId = i + 1;
				const restartWorker = (code?: number, signal?: string) => {
					if (signal != null) {
						serviceLogger.crit(
							`worker-${workerId} killed with signal ${signal}`,
						);
					}
					if (code != null) {
						serviceLogger.crit(`worker-${workerId} exited with code ${code}`);
					}
					if (terminating) {
						// Don't restart the work if we're terminating
						return;
					}
					const env = {
						...process.env,
						WORKER_ID: workerId,
						SERVICE_ID: serviceInstance.getId(),
						VPN_VERBOSE_LOGS: verbose,
					};
					cluster.fork(env).on('exit', restartWorker);
				};
				restartWorker();
			});

Page- (approved):

PR #275, opened 2022-07-08
Use a rust binary rather than bash for handling openvpn auth

Page-:
Change-type: patch

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---


auth/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+    let auth_control_file = std::env::var("auth_control_file").unwrap();
+    let vpn_api_port = std::env::args().nth(1).unwrap();
+
+    match ureq::post(&format!("http://127.0.0.1:{}/api/v1/auth/", vpn_api_port))
+        .set("Content-type", "application/json")
+        .send_string(&format!(
+            "{{\"username\":\"{}\",\"password\":\"{}\"}}",
+            username, password
+        )) {
+        Ok(_) => {
```
dfunckt, line 32: Semantically this is different than before: previously it would take as “success” the presence of “OK” into the body, now it’ll consider success (I would imagine) any 2xx response code.

Page-: Yeah it's technically slightly differently but the main reason for checking `OK` is that in bash it's easier to check the response body than the status code. Either way for our purposes the status code is perfectly ok to check since it's either `200 OK` or `400` or `401`

PR #281, opened 2022-08-10
Allow specifying an internal hostname for balena-api comms

Page-:
Change-type: minor

Depends on https://github.com/balena-io/open-balena-api/pull/1086


src/utils/clients.ts
```diff
@@ -65,7 +65,7 @@ export const setConnected = (() => {
 				try {
 					const response: Response = await request
 						.post({
-							url: `https://${BALENA_API_HOST}/services/vpn/client-${eventType}`,
+							url: `${BALENA_API_INTERNAL_HOST}/services/vpn/client-${eventType}`,
```
dfunckt, line 68: Do you not need to set the protocol too? If it's implied and defaults to `http` I'd rather it was made explicit.

dfunckt: Never mind, I just noticed protocol is included in the env var value, set a little further down.

Page-: Yup it's set in `config.ts` rather than hardcoding the `https`/`http` in the rest of the code

src/utils/config.ts
```diff
@@ -125,7 +156,15 @@ export const VPN_BASE_MANAGEMENT_PORT = intVar('VPN_BASE_MANAGEMENT_PORT');
...
 export const VPN_BYTECOUNT_INTERVAL = intVar('VPN_BYTECOUNT_INTERVAL', 0);
 
-export const BALENA_API_HOST = requiredVar('BALENA_API_HOST');
+const apiHostForInternalUse = requiredMultiVar(
+	'BALENA_API_INTERNAL_HOST',
+	'BALENA_API_HOST',
+);
+// If we're using a dedicated internal host then we use http, if it's a shared external one it needs to be https
+export const BALENA_API_INTERNAL_HOST =
+	apiHostForInternalUse === optionalVar('IMAGE_MAKER_INTERNAL_HOST')
```
fisehara, line 165: Shouldn't this be: `BALENA_API_INTERNAL_HOST`

``` suggestion
apiHostForInternalUse === optionalVar('BALENA_API_INTERNAL_HOST')
```

fisehara, line 165: Shouldn't this be: `BALENA_API_INTERNAL_HOST`

``` suggestion
apiHostForInternalUse === optionalVar('BALENA_API_INTERNAL_HOST')
```

fisehara (commented):
IDK where this `IMAGE_MAKER_INTERNAL_HOST` is coming from.

ab77 (approved):

PR #285, opened 2022-08-15
Always send a connect event to the API if the connected worker changes

Page-:
This solves an issue that could happen where the device reconnects in
rapid succession and connects to another VPN container and then back to
the current one without yet having been recognized as disconnected from
the current VPN container and as such we would not send a connect event
for the reconnection. This would result in the device appearing to be
offline for the API (because it disconnected from the other instance)
whilst actually still being connected, just to this instance

Change-type: patch

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -68,7 +68,7 @@ const checkDeviceAuth = memoize(
 export const apiFactory = (serviceId: number) => {
 	const api = express.Router();
 
-	const workerMap: _.Dictionary<string> = {};
+	const workerMap: _.Dictionary<number> = {};
 
 	const logger = getLogger('vpn', serviceId);
 
@@ -84,14 +84,14 @@ export const apiFactory = (serviceId: number) => {
 		metrics.inc(Metrics.OnlineDevices);
 		metrics.inc(Metrics.TotalDevices);
 
-		const workerId = req.params.worker;
+		const workerId = parseInt(req.params.worker, 10);
 		const uuid = req.body.common_name;
-		if (workerMap[uuid] != null && workerMap[uuid] !== req.params.worker) {
+		if (workerMap[uuid] != null && workerMap[uuid] !== workerId) {
 			metrics.dec(Metrics.OnlineDevices);
 		}
 
 		workerMap[uuid] = workerId;
-		clients.setConnected(uuid, serviceId, true, logger);
+		clients.setConnected(uuid, serviceId, workerId, true, logger);
 	});
 
 	api.post('/api/v1/auth/', fromLocalHost, async function (req, res) {
@@ -137,7 +137,7 @@ export const apiFactory = (serviceId: number) => {
 			metrics.histogram(Metrics.SessionDuration, req.body.time_duration);
 		}
 
-		const workerId = req.params.worker;
+		const workerId = parseInt(req.params.worker, 10);
 		const uuid = req.body.common_name;
 
 		if (workerMap[uuid] !== workerId) {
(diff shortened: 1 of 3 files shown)


Page-:
@balena-ci I self-certify!

src/vpn-worker.ts
```diff
@@ -127,8 +127,6 @@ const worker = async (instanceId: number, serviceId: number) => {
 					const cn = client.uuid;
 					try {
 						logger.info(`disconnecting ${cn}`);
-						// update device state
-						clients.setConnected(cn, serviceId, false, logger);
```
Page-, line 131: I removed this because disconnecting the device from the VPN should trigger the natural disconnect tracking method

src/api.ts
```diff
@@ -84,14 +84,14 @@ export const apiFactory = (serviceId: number) => {
...
+		const workerId = parseInt(req.params.worker, 10);
 		const uuid = req.body.common_name;
-		if (workerMap[uuid] != null && workerMap[uuid] !== req.params.worker) {
+		if (workerMap[uuid] != null && workerMap[uuid] !== workerId) {
 			metrics.dec(Metrics.OnlineDevices);
 		}
 
 		workerMap[uuid] = workerId;
-		clients.setConnected(uuid, serviceId, true, logger);
+		clients.setConnected(uuid, serviceId, workerId, true, logger);
```
pranavpeshwe, line 94: Can we add a log statement here that logs uuid, serviceId & workerId? Will be useful for debugging these kind of issues.

src/api.ts
```diff
@@ -137,7 +137,7 @@ export const apiFactory = (serviceId: number) => {
 			metrics.histogram(Metrics.SessionDuration, req.body.time_duration);
 		}
 
-		const workerId = req.params.worker;
+		const workerId = parseInt(req.params.worker, 10);
 		const uuid = req.body.common_name;
 
 		if (workerMap[uuid] !== workerId) {
```
pranavpeshwe, line 143: Will this `if()` ever be true now? In what case?

src/api.ts
```diff
@@ -157,7 +157,7 @@ export const apiFactory = (serviceId: number) => {
 
 		metrics.dec(Metrics.OnlineDevices);
 
-		clients.setConnected(uuid, serviceId, false, logger);
+		clients.setConnected(uuid, serviceId, workerId, false, logger);
```
pranavpeshwe, line 160: A debug log here would also help I believe.

src/utils/clients.ts
```diff
@@ -81,6 +83,11 @@ export const setConnected = (() => {
 					// Update the current state on success
 					for (const uuid of uuidChunk) {
 						deviceStates[uuid].currentConnected = connected;
+						// We mark the currentWorkerId to match the target since the key is that we have sent a new connect
+						// event since the worker has connected in case it went from eg `us-1` -> `other-1` -> `us-2` and
+						// `other` is seen as being in charge of the device when it actually should be us
+						deviceStates[uuid].currentWorkerId =
+							deviceStates[uuid].targetWorkerId;
 						logger.debug(
 							`successfully updated state for device: uuid=${uuid} connected=${connected}`,
```
pranavpeshwe, line 92: Would be good to have `deviceStates[uuid].currentWorkerId` logged.

src/vpn-worker.ts
pranavpeshwe: I wonder why this had been added. To expedite the state transition for some reason?

PR #291, opened 2022-08-25
Add VPN_SHAPER_BYTES_PER_SECOND config

Page-:
Change-type: minor

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---


dfunckt:
I'm having a look around the web and it appears that `shaper` is not compatible with `mode server` and that openVPN does not actually support traffic shaping per connection from the server's side. The references I found are quite old though -- more than 5 years old the most recent one.

Page-:
Yeah I'm still investigating that @dfunckt, because we also use the `topology p2p`

Page-:
Yup, it's not going to work

PR #292, opened 2022-08-29
Avoid delays in client disconnection from affecting the draining cadence

Page-:
This avoids situations where due to client disconnection delays we end
up with clients being disconnected at a slower cadence than expected
and meaning that there may still be significant numbers connected at
the point the SIGTERM timeout is reached

Change-type: patch

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---

### DIFF
--- a/src/vpn-worker.ts
+++ b/src/vpn-worker.ts
@@ -122,19 +122,17 @@ const worker = async (instanceId: number, serviceId: number) => {
 				`disconnecting ${clientCount} clients, spaced by ${delayMs}ms`,
 			);
 			let timeToKill = DEFAULT_SIGTERM_TIMEOUT;
-			for (const client of drainQueue) {
-				try {
-					const cn = client.uuid;
+			for (const { uuid: cn } of drainQueue) {
+				// Trigger the disconnecting the client in the background so as to avoid it delaying
+				// the overall cadence of disconnections but whilst still having error handling for it
+				(async () => {
 					try {
 						logger.info(`disconnecting ${cn}`);
-						// disconnect client from VPN
 						await vpn.killClient(cn);
 					} catch (err) {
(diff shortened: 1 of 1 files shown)


fisehara (approved):
LGTM

src/vpn-worker.ts
```diff
@@ -122,19 +122,17 @@ const worker = async (instanceId: number, serviceId: number) => {
...
-						// disconnect client from VPN
 						await vpn.killClient(cn);
 					} catch (err) {
 						logger.warning(`${err} while killing ${cn} on worker ${serviceId}`);
 					}
-				} catch (err) {
-					logger.warning(`received '${err}' trying to disconnect client`);
-				}
+				})();
 				timeToKill = timeToKill - delayMs;
```
PranavPeshwe, line 136: Hey @Page- , @fisehara , Now that we are using an async IIFE, shouldn't all the logic to wait between kill requests be removed? I do not know why @ab77 decided to introduce a delay between kills; there must have been some reason I suppose.
Hope I'm not missing something obvious here 🤔

fisehara: @PranavPeshwe from my understanding, the drain timeout is still a valid approach to avoid a vpn.killClient() rush.
But therefore, I don't know enough about the vpn.killClient().
What I understand is, that disconnecting a client may take some time but we don't want to wait until the client has been disconnected to kill the next client, but we want ensure that we don't call too many killClient()

PranavPeshwe: @fisehara , Thanks! You are correct 👍🏻 The `await` for the timeout will make the async IIFE wait. I had missed it. And regarding the delay between kills - probably it was intended to limit the rate of `exec` calls on the pod.

PR #296, opened 2022-09-07
Set desired es-version for downstream modules that support it

Page-:
Notably this will cause pinejs-client to use a version with native async/await and hence allow for async stack traces

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #297, opened 2022-09-07
Stop automatic int parsing of openvpn events to avoid false positives

Page-:
This avoids the case where a uuid could be a valid int and be parsed as a false positive and cause issues. It also avoids the overhead of parsing ints when their parsed form may never be needed.

Change-type: patch


Page-:
@balena-ci I self-certify!

PR #314, opened 2023-01-18
Push ping-exit directive to clients rather than ping-restart

klutchell:
balenaOS devices drop privileges after starting OpenVPN so the SIGUSR1 is unable to recreate the socket.

Instead, have the server push the ping-exit directive as per the OpenVPN docs. This way balenaOS will stop and restart the service.

Change-type: minor
Signed-off-by: Kyle Harding <kyle@balena.io>
See: https://github.com/balena-os/meta-balena/issues/1779

---
##### Contributor checklist

- [x] Tests are included

##### Reviewer Guidelines
---

### DIFF
--- a/config/confd/templates/server.conf.tmpl
+++ b/config/confd/templates/server.conf.tmpl
@@ -11,7 +11,12 @@ persist-tun
 
 proto tcp4-server
 compress lzo
-keepalive {{ getenv "VPN_KEEPALIVE_FREQUENCY" "10" }} {{ getenv "VPN_KEEPALIVE_TIMEOUT" "60" }}
+
+# Tell clients to exit on keepalive timeouts
+ping {{ getenv "VPN_KEEPALIVE_FREQUENCY" "10" }}                    # interval
+ping-exit {{ (atoi (getenv "VPN_KEEPALIVE_TIMEOUT" "60")) mul 2 }}  # timeout*2
+push "ping {{ getenv "VPN_KEEPALIVE_FREQUENCY" "10" }}"             # interval
+push "ping-exit {{ getenv "VPN_KEEPALIVE_TIMEOUT" "60" }}"          # timeout
 
 # Increase the max clients from the default of 1024 (hard limit applied by subnet/pool config)
 max-clients 32768


klutchell:
The CircleCI test logs show the 'ping-exit' directives we were hoping for.
```
notice: [vpn-821.1] user5/127.0.0.1:56904 SENT CONTROL [user5]: 'PUSH_REPLY,ping 10,ping-exit 60,sndbuf 0,rcvbuf 0,route 100.64.0.1,socket-flags TCP_NODELAY,ifconfig 100.64.0.6 100.64.0.1,peer-id 0,cipher AES-128-GCM' (status=1)
```

jakogut:
@klutchell This looks promising, I've just identified another support thread where a user was affected by this issue. Is there anything else this PR needs to be ready for review?

klutchell:
Waiting on https://github.com/balena-io/open-balena-vpn/pull/316 to unblock the tests.

config/confd/templates/server.conf.tmpl
```diff
@@ -11,7 +11,12 @@ persist-tun
 
 proto tcp4-server
 compress lzo
-keepalive {{ getenv "VPN_KEEPALIVE_FREQUENCY" "10" }} {{ getenv "VPN_KEEPALIVE_TIMEOUT" "60" }}
+
+# Tell clients to exit on keepalive timeouts
+ping {{ getenv "VPN_KEEPALIVE_FREQUENCY" "10" }}                    # interval
+ping-exit {{ (atoi (getenv "VPN_KEEPALIVE_TIMEOUT" "60")) mul 2 }}  # timeout*2
```
klutchell, line 17: Do we actually want/need `ping-exit` on the server side? I know that `keepalive` was setting `ping-restart` but I couldn't tell from the docs if we actually want the server to use `ping-exit` or stick with `ping-restart` or neither?

config/confd/templates/server.conf.tmpl
```diff
@@ -11,7 +11,12 @@ persist-tun
...
+push "ping {{ getenv "VPN_KEEPALIVE_FREQUENCY" "10" }}"             # interval
+push "ping-exit {{ getenv "VPN_KEEPALIVE_TIMEOUT" "60" }}"          # timeout
```
klutchell, line 19: I have confirmed these are received by the client.

config/confd/templates/server.conf.tmpl
Page-: I think both should match

majorz: The docs mention how `ping-restart` is handled on the server side:

"In server mode, --ping-restart, --inactive or any other type of internally generated signal will always be applied to individual client instance objects, never to whole server itself. Note also in server mode that any internally generated signal which would normally cause a restart, will cause the deletion of the client instance object instead." 

However nothing is being mentioned for `ping-exit` for the server mode. My gut feeling is that we need to keep `ping-restart` on the server side.

majorz: `ping-exit` on the server may indeed have undesired behavior on the server side: https://docs.netgate.com/pfsense/en/latest/vpn/openvpn/configure-server-ping.html#ping-method

klutchell:
I self-certify!

PR #330, opened 2023-04-21
Add a configurable (10s default) delay on auth failure

Page-:
This allows delaying returning an auth failure which will also delay the device retrying

Change-type: minor

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---


src/api.ts
```diff
@@ -126,6 +128,7 @@ export const apiFactory = (serviceId: number) => {
 				metrics.inc(Metrics.AuthFailuresByUuid, undefined, {
 					device_uuid: req.body.common_name,
 				});
+				await setTimeout(DELAY_ON_AUTH_FAIL)
```
dfunckt, line 131:
```suggestion
				await setTimeout(DELAY_ON_AUTH_FAIL);
```

Page-: Yeah, the problem is that husky was updated to 8.x in this repo without updating to 8.x config - I've fixed that now

dfunckt (approved):

PR #360, opened 2023-08-14
shutdown dependency between systemd units

ab77:


config/services/open-balena-vpn.service
```diff
@@ -26,17 +27,9 @@ WorkingDirectory=/usr/src/app
...
-# https://www.freedesktop.org/software/systemd/man/systemd.service.html#TimeoutStopSec=
-# ensure k8s/terminationGracePeriodSeconds matches
-TimeoutStopSec=${DEFAULT_SIGTERM_TIMEOUT}
 
 Restart=always
 SyslogIdentifier=vpn
 
 [Install]
-WantedBy=basic.target
+WantedBy=multi-user.target
```
Page-, line 35: Does this work with `multi-user.target` again? I know I had to change it to fix things for debian bullseye

PR #406, opened 2024-03-05
allow OpenVPN carrier protocol configuration

ab77:
Requires:
* https://github.com/balena-io/open-balena-api/pull/1570
* https://github.com/balena-io/environment-staging/pull/3144


src/utils/config.ts
```diff
@@ -70,6 +70,7 @@ export const MAXIMUM_DRAIN_DELAY = Math.round(
 	(1 * MINUTES) / intVar('MINIMUM_DRAIN_RATE_PER_MINUTE', 500),
 );
 
+// must be max(1) if VPN_OPENVPN_PROTO =~ udp
 export const VPN_INSTANCE_COUNT = getInstanceCount('VPN_INSTANCE_COUNT');
```
thgreasi, line 74: Should we actually add code for this?
```suggestion
export const VPN_INSTANCE_COUNT = getInstanceCount('VPN_INSTANCE_COUNT');
if (optionalVar('VPN_OPENVPN_PROTO') === 'udp' && VPN_INSTANCE_COUNT > 1) {
	throw new Error('VPN_INSTANCE_COUNT can't be > 1 when VPN_OPENVPN_PROTO is udp');
}
```
or
```suggestion
export let VPN_INSTANCE_COUNT = getInstanceCount('VPN_INSTANCE_COUNT');
if (optionalVar('VPN_OPENVPN_PROTO') === 'udp') {
	console.error('VPN_INSTANCE_COUNT can't be > 1 when VPN_OPENVPN_PROTO is udp! Falling back to 1.');
	VPN_INSTANCE_COUNT = 1;
}
```

config/confd/templates/server.conf.tmpl
```diff
@@ -57,3 +57,8 @@ tcp-nodelay
 # disable tls renegotiation hourly in favour of every 1GB to avoid load spikes
 reneg-sec 0
 reneg-bytes 1073741824
+
+{{ if getenv "VPN_UDP_SERVER_EXPLICIT_EXIT_NOTIFY" }}
+# send RESTART control channel command to connected clients
+{{ getenv "VPN_UDP_SERVER_EXPLICIT_EXIT_NOTIFY" }}
```
Page-, line 63: What is meant to be in here? This seems like it could contain any openvpn config?

ab77: @Page- it is meant for controlling this setting:

```
--explicit-exit-notify n
--
  | In UDP client mode or point-to-point mode, send server/peer an exit notification if tunnel is restarted or OpenVPN process is exited. In client mode, on exit/restart, this option will tell the server to immediately close its client instance object rather than waiting for a timeout.The n parameter (default 1 if not present) controls the maximum number of attempts that the client will try to resend the exit notification message.In UDP server mode, send RESTART control channel command to connected clients. The n parameter (default 1 if not present) controls client behavior. With n = 1 client will attempt to reconnect to the same server, with n = 2 client will advance to the next server.OpenVPN will not send any exit notifications unless this option is enabled.
```

I suppose we can just use this parameter for specifying `n` only, so it would become:

```
{{ if getenv "VPN_UDP_SERVER_EXPLICIT_EXIT_NOTIFY" }}
# send RESTART control channel command to connected clients
explicit-exit-notify {{ getenv "VPN_UDP_SERVER_EXPLICIT_EXIT_NOTIFY" }}
{{ end }}
```

src/utils/config.ts
```diff
@@ -70,7 +70,19 @@ export const MAXIMUM_DRAIN_DELAY = Math.round(
 	(1 * MINUTES) / intVar('MINIMUM_DRAIN_RATE_PER_MINUTE', 500),
 );
 
-export const VPN_INSTANCE_COUNT = getInstanceCount('VPN_INSTANCE_COUNT');
+export let VPN_INSTANCE_COUNT = getInstanceCount('VPN_INSTANCE_COUNT');
```
klutchell, line 73: my understanding is that this export will not update if the value changes below, but someone else could confirm (or we could write a small test)

Page-: It should be fine, especially with ESM as open-balena-vpn uses (but even with typescript+commonjs it should be ok due to how it transpiles, it's plain commonjs that could be problematic in some cases)

src/utils/config.ts
```diff
@@ -70,7 +70,19 @@ export const MAXIMUM_DRAIN_DELAY = Math.round(
...
+const VPN_OPENVPN_PROTO = optionalVar('VPN_OPENVPN_PROTO');
+if (VPN_OPENVPN_PROTO) {
+	if (/^udp[4-6]?$/.test(VPN_OPENVPN_PROTO)) {
+		if (VPN_INSTANCE_COUNT > 1) {
+			console.error(
```
Page-, line 78: I wonder if it would be better for this to just straight up error rather than quietly falling back to single process, because if you're expecting eg 4 processes per instance and you only get 1 process per instance you might hit scaling issues and might not immediately realize if you look at cpu stats and see eg 25% cpu usage that it's because there's only 1 process and not the 4 you expected.

Page- (approved):

PR #427, opened 2024-05-28
capture stderr logs from client connect/disconnect script

ab77:
* this script signals the API that a client has connected (or disconnected) and it is important to understand if any of these events are failing to reach open-balena-vpn API as they can lead to balena-api being unaware of client VPN state change

change-type: patch

### DIFF
--- a/openvpn/scripts/client-connect-disconnect.sh
+++ b/openvpn/scripts/client-connect-disconnect.sh
@@ -17,7 +17,7 @@
 
 function cleanup() {
 	if [[ $? -gt 0 ]]; then
- 		cat </tmp/$PPID.tmp >/tmp/stderr-$PPID.log  		
+ 		cat </tmp/$PPID.tmp >>/tmp/stderr-$PPID.log  		
 	fi
 	rm /tmp/$PPID.tmp
 }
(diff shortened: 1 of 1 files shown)


openvpn/scripts/client-connect-disconnect.sh
```diff
@@ -15,17 +15,25 @@
 # You should have received a copy of the GNU Affero General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+function cleanup() {
+	if [[ $? -gt 0 ]]; then
+ 		cat </tmp/$PPID.tmp >/tmp/stderr-$PPID.log
```
joshbwlng, line 20: Would it be better to append to the file instead of overwrite it? (Same question for `2>/tmp/$PPID.tmp` redirects below.)

ab77: Good point, do you think we could have PID collisions here?

joshbwlng: I honestly don't know, but my thinking is that we'd probably rather have more verbose logs and not accidentally overwrite something possibly valuable.

joshbwlng (approved):

PR #590, opened 2025-09-24
Restore nproc environment variable for confd templates

klutchell:
This env var was previously provided by a confd systemd overlay and never replaced when we moved to s6.

As a result, haproxy was running only 1 vpn worker instead of 16.

Change-type: patch
See: https://balena.fibery.io/Work/Project/Make-Cloudlink-deployable-again-1909

https://github.com/balena-io/open-balena-vpn/blob/dac57011ec9eec3bb0e99d54867d49204d47369e/config/services/confd.service.d/confd.conf#L17

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -49,6 +49,13 @@ services:
         s6-svc -d /run/service/open-balena-vpn
 
         npm install
+
+        mkdir -p /dev/net /run/openvpn /run/openvpn-client /run/openvpn-server
+
+        if [ ! -c /dev/net/tun ]; then
+          mknod /dev/net/tun c 10 200
+        fi
+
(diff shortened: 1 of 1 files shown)


config/s6-overlay/scripts/entry.sh
```diff
@@ -2,6 +2,11 @@
 
 set -e
 
+# Inject number of processors into environment for confd.
+# This value is used by haproxy and openvpn config templates.
+NPROC=$(nproc)
+export NPROC
```
klutchell, line 8: This doesn't seem to be getting picked up by confd when run again as supervised which is odd, will need to do more debugging.

PR #593, opened 2025-09-26
Move setting haproxy to drain mode logic from bash into nodejs

Page-:
Change-type: patch

Connects-to: # 
See: <url> 
Depends-on: <url> 

---
##### Contributor checklist

- [ ] Tests are included

##### Reviewer Guidelines
---


src/vpn-worker.ts
```diff
@@ -113,6 +113,10 @@ const worker = async (instanceId: number, serviceId: number) => {
 	});
 
 	const drainConnections = _.once(async () => {
+		logger.info(`setting haproxy for '${instanceId}' into drain mode`);
+		await new HAProxy('/var/run/haproxy.sock').drain(
+			`vpn-workers/vpn${instanceId}`,
+		);
```
klutchell, line 119: That is... so much cleaner 🎉

docker-compose.test.yml
```diff
@@ -56,7 +57,7 @@ services:
           mknod /dev/net/tun c 10 200
         fi
 
-        /command/with-contenv npm run test-unit
+        # /command/with-contenv npm run test-unit
```
klutchell, line 60: Why are we skipping this test?

Page-: It was a local change that wasn't meant to be included, but basically when I was debugging the second test command I didn't want to wait ages for this one each time

PR #636, opened 2025-12-30
Drop the use of classids files

jaomaloy:
We base the classid needed for tc rules
off of IP (last 2 octet) hash instead of
writing to a file everytime a device connects.

Change-type: patch
Fibery: https://balena.fibery.io/Work/Project/(Cloudlink)-Monitor-and-ready-Bandwidth-limiting-and-openVPN-exporter-for-production-1731

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,12 +82,6 @@ Both parameters must be specified to enable throttling. Rate formats supported:
 
 For troubleshooting throttling issues:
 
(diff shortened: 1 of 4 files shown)


ab77:
@jaomaloy we may need a sync call to review, as I've lost all the context for this.

jaomaloy:
@ab77 would love to, I'll find a slot next week!

ab77 (approved):

PR #680, opened 2026-04-21
Define session_duration in the primary instead of workers

jaomaloy:
The vpn_session_duration histogram was using default Prometheus
buckets instead of the configured duration buckets - because
this was in the worker configuration/definition instead of
the primary where it is called. Also, update bucket boundaries
to meaningful durations for VPN sessions.

Change-type: patch
FIbery: https://balena.fibery.io/Work/Project/Data-Infrastructure-for-Usable-Time-Series-of-Online-Devices-1879


ab77:
@jaomaloy we may need a sync call to review, as I've lost all the context for this.

ab77 (approved):

